PR #3, opened 2017-09-21
Add tuple type

mame:
What do you think?  Note that this pullreq changes the type of `[1, 2]` from `Array<Integer>` to `[Integer, Integer]`.


lib/steep/type_assignability.rb
```diff
@@ -55,6 +55,16 @@ def test(src:, dest:, known_pairs: [])
         known_pairs.include?([src, dest])
       when src.is_a?(Types::Name) && dest.is_a?(Types::Name)
         test_interface(resolve_interface(src.name, src.params), resolve_interface(dest.name, dest.params), known_pairs)
+      when src.is_a?(Types::Tuple) && dest.is_a?(Types::Tuple)
+        src.types.zip(dest.types) do |type1, type2|
+          test(src: type1, dest: type2, known_pairs: known_pairs)
+        end
+      when src.is_a?(Types::Name) && dest.is_a?(Types::Tuple)
+        params = [dest.types.uniq.size == 1 ? dest.types.first : Types::Any.new]
```
soutaro: Probably, we can have union type (but the array type testing is broken still).

lib/steep/types/tuple.rb
```diff
@@ -0,0 +1,53 @@
...
+      # @type const TypeName::Instance: Steep__TypeName.class
+
+      # @dynamic types
+      attr_reader :types
+
+      def initialize(types:)
+        @types = types
+      end
+
+      def self.interface(types: [])
```
soutaro: I don't think we have to have these factory methods. (I had implemented these for `Name` type.)

lib/steep/type_assignability.rb
```diff
@@ -55,6 +55,16 @@ def test(src:, dest:, known_pairs: [])
         known_pairs.include?([src, dest])
       when src.is_a?(Types::Name) && dest.is_a?(Types::Name)
         test_interface(resolve_interface(src.name, src.params), resolve_interface(dest.name, dest.params), known_pairs)
+      when src.is_a?(Types::Tuple) && dest.is_a?(Types::Tuple)
+        src.types.zip(dest.types) do |type1, type2|
+          test(src: type1, dest: type2, known_pairs: known_pairs)
```
soutaro: Should we have covariant array? (I'm not how to do for this.)

soutaro:
Thanks, this is awesome!

I think there are two things we should consider:

* I want to use union types when casting tuple to array (but it does not make sense because of a bug #5)
* I don't have a good opinion about having covariant tuple (Making invariant currently should be the safe option?)

mame:
Thank you for the review.

> I want to use union types when casting tuple to array

Do you mean, for example, assigning `Tuple<Integer, Integer> | Array<Integer>` for `[1, 1]`?  I never thought about that, but it sounds like a good idea.  I'll try later.

> I don't have a good opinion about having covariant tuple

It is hard to decide...  I think that a tuple-like array is normally used as an immutable, but it is indeed mutable.  Adding both `ImmutableTuple` and `MutableTuple` may be a possible option, but I'm unsure if this is a good idea.

soutaro:
> I mean something like `[Integer, String]` <: `Array<String|Integer>`

Maybe confusing... I expect `[1, "string"][i]` to have `Integer | String` instead of `any`.

PR #5, opened 2017-09-23
Fix parametrized types compatibility test

### LINKED ISSUE #4: `Array<Integer>` and `Array<String>` are assignable
If two `name`s are identical, `TypeAssignability#test_interface` returns true with no check of method signatures.  Steep says nothing for the following program.  Is this intended?

```
# @type var a: Array<Integer>
# @type var b: Array<String>

a = []
b = a
a = b
```

soutaro:
This PR is to fix #4.

### DIFF
--- a/stdlib/builtin.rbi
+++ b/stdlib/builtin.rbi
@@ -49,7 +49,7 @@ class Array<'a>
   def join: (any) -> String
   def all?: { (any) -> any } -> _Boolean
   def sort_by: { ('a) -> any } -> Array<'a>
-  def zip: <'b> (Array<'b>) -> Array<Array<'a, 'b> >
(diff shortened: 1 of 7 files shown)


stdlib/builtin.rbi
```diff
@@ -49,7 +49,7 @@ class Array<'a>
   def join: (any) -> String
   def all?: { (any) -> any } -> _Boolean
   def sort_by: { ('a) -> any } -> Array<'a>
-  def zip: <'b> (Array<'b>) -> Array<Array<'a, 'b> >
+  def zip: <'b> (Array<'b>) -> Array<any>
```
soutaro: Non-terminating was caused by this invalid `Array` type application 😢

PR #16, opened 2018-02-26
Support Ruby 2.5

pocke:
I guess steep support only Ruby 2.5, because steep uses `yield_self`, and it's tested only Ruby 2.5 on Travis CI.


soutaro:
Thanks!

PR #18, opened 2018-02-27
Add support error handling sytnax (`begin - rescue - end`)

pocke:
This pull-request makes Steep to support `begin - rescue - end` syntax.

### DIFF
--- a/lib/steep/type_construction.rb
+++ b/lib/steep/type_construction.rb
@@ -282,7 +282,7 @@ module Steep
       Steep.logger.tagged "synthesize:(#{node.location.expression.to_s.split(/:/, 2).last})" do
         Steep.logger.debug node.type
         case node.type
-        when :begin
+        when :begin, :kwbegin
           yield_self do
             type = each_child_node(node).map do |child|
               synthesize(child)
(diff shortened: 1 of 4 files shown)


lib/steep/type_construction.rb
```diff
@@ -697,6 +697,36 @@ def synthesize(node)
 
           typing.add_typing(node, union_type(*types))
 
+        when :rescue
+          yield_self do
+            body, *resbodies, else_node = node.children
+            body_type = synthesize(body) if body
+            resbody_types = resbodies.map { |resbody| synthesize(resbody) }
+            else_type = synthesize(else_node) if else_node
```
pocke: When `begin` has an `else`, `begin` body value is always ignored.
For example: `begin 1 else 2 end == 2 # => true`

lib/steep/type_construction.rb
```diff
@@ -697,6 +697,36 @@ def synthesize(node)
...
+            types.concat resbody_types
+            types << else_type
+            typing.add_typing(node, union_type(*types))
+          end
+
+        when :resbody
+          yield_self do
+            klasses, asgn, body = node.children
+            synthesize(klasses) if klasses
+            synthesize(asgn) if asgn
```
pocke: I'm not sure what should Steep decide assigned type. So currently the `synthesize(asgn)` is always fallback any.


I guess Steep can decide the type in limited cases. 
For example:

```ruby
begin
rescue => e # The `e` is a ::StandardError always
end

begin
rescue ::StandardError => e # # The `e` is also a ::StandardError always
end

begin
rescue StandardError => e # Steep cannot decide type.
end
```

But I'm not sure how to implement it. Can you help me?

lib/steep/type_construction.rb
```diff
@@ -697,6 +697,36 @@ def synthesize(node)
...
           typing.add_typing(node, union_type(*types))
 
+        when :rescue
+          yield_self do
+            body, *resbodies, else_node = node.children
+            body_type = synthesize(body) if body
+            resbody_types = resbodies.map { |resbody| synthesize(resbody) }
+            else_type = synthesize(else_node) if else_node
+            types = []
+            types << body_type unless else_node
```
pocke: When begin has an else, begin body value is always ignored.
For example: begin 1 else 2 end == 2 # => true

lib/steep/type_construction.rb
soutaro: No implementation we can. The example above assumes `===` implementation, which Steep cannot (or, does not) for now.

This is the correct implementation. 🥁

soutaro:
I will take a look at this PR soon, probably tomorrow. 👀

lib/steep/type_construction.rb
soutaro: My understanding was incorrect... It seems that `rescue` clause does not use `===`. However, current implementation still looks ok because Steep does not have *branch local type annotation* yet. So, I'm merging this PR.

pocke:
Thank you!

PR #31, opened 2018-05-23
Add a simple type guesser into scaffold

mame:
scaffold guesses a type of constants, instance variables, and return
values of a method by investigating the syntax tree of left-hand
expression or method body.

```
class Foo
  CONST = 1              #=> Foo:CONST: Integer
  def func
    @ivar = "Str"        #=> @ivar: String
    cond? ? :foo : :bar  #=> def func: () -> Symbol
  end
end
```


soutaro:
👏🙈👻

PR #32, opened 2018-05-25
add rubocop gem for refactoring.

holylance:
In my case, I always use the rubocop gem in my project.
It is a very cool gem for refactoring such as find useless space.

It is very easy to install and run a test.
`
> bundle
> rubocop
126 files inspected, 1977 offenses detected
`

I found 1977 offenses with this gem in your project.
How about your thinking?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 /tmp/
 /lib/steep/parser.output
 /lib/steep/parser.rb
+/.ruby-version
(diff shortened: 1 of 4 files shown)


soutaro:
Thank you very much for sending a pull request.

However, I don't think we should setup RuboCop for this project, at least for now.
I know what RuboCop is and how that tool is useful. But I don't want to use that for this project because I can review all of the pull requests still (it's not so much).

Thanks again, but I'm closing this.

Issue #42, opened 2018-07-03
What is this project's licence?


ned:
This looks like a great project, but I couldn't see any licensing info. Are there any plans to open source the code?

soutaro:
It's not intentional. 😱 I feel the project is open source but I found I haven't declared that.

Thanks!

ned:
Awesome, thanks for adding the license!

It's probably worth contacting other contributors to make sure they don't object to the licence

ybiquitous:
@soutaro The commit 3f2d6db86187027df6f75c502003d2c95babe6e5 seems to fix this issue, so do you close this?

PR #43, opened 2018-07-03
Steep is licensed under the MIT license

soutaro:
#42 🎉


soutaro:
@stereobooster @pocke @mame @hanachin  Hi contributors! I have declared that the license for this project is MIT. Please tell me if you have any other opinion about this. (Even though it's already merged, I can revert if we need. 😎)

mame:
I agree with MIT.

hanachin:
I agree with MIT.

PR #56, opened 2018-10-09
Add `steep version` subcommand

ybiquitous:
This PR adds a new `steep version` subcommand, which only outputs the Steep version to STDOUT.

E.g.

```
$ steep version
0.7.0
```

### DIFF
--- a/lib/steep/cli.rb
+++ b/lib/steep/cli.rb
@@ -16,7 +16,7 @@ module Steep
     end
 
     def self.available_commands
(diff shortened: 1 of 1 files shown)


koic:
IMO, About the command version I think that it is user-friendly to replace it with an option (`steep --version` or `steep -v`) from a subcommand (`steep version`).

ybiquitous:
@koic Your idea seems good to me. I will try to implement `--version` / `-v` options. 👍

soutaro:
I like the `version` sub command in fact. 😄

PR #57, opened 2018-10-09
Add option to show Steep's version

ybiquitous:
This PR adds a new option `--version` (`-v`), which only outputs the Steep version to STDOUT.

E.g.

```
$ steep --version
steep 0.7.0

$ steep -v
steep 0.7.0
```

### DIFF
--- a/lib/steep/cli.rb
+++ b/lib/steep/cli.rb
@@ -2,6 +2,8 @@ require 'optparse'
 
 module Steep
   class CLI
+    ::Version = Steep::VERSION
+
     attr_reader :argv
     attr_reader :stdout
     attr_reader :stdin
@@ -20,6 +22,12 @@ module Steep
     end
 
     def setup_global_options
+      version = OptionParser.new.version
+      if version
+        stdout.puts version
+        return false
+      end
+
       true
     end
 


ybiquitous:
This may conflict with PR #56.

lib/steep/cli.rb
```diff
@@ -20,6 +20,10 @@ def self.available_commands
     end
 
     def setup_global_options
+      OptionParser.new do |opts|
+        opts.version = Steep::VERSION
+      end.parse(argv)
+
```
ybiquitous: I think this is a minimal implementation, but it is confusing me the behavior which `OptionParser#parse` exits the program when it find a `--verison` option. 🤔

ybiquitous:
Actually, I don't know which is better with or without the program name (`steep`). 😕 

##### With the name (`OptionParser` default behavior):

```
$ steep --version
steep 0.7.0
```

##### Without the name:

```
$ steep --version
0.7.0
```

For some users who want to parse STDOUT and get Steep's version, the latter may be better... 🤔

lib/steep/cli.rb
soutaro: I think defining `::Version` is better to handle other options in sub commands.

soutaro:
We can change the behavior later if we find any problem. 😆

soutaro:
Because optparse accepts `--version` option, we should support that too. Defining `::Version` constant seems the best solution for this purpose to me.

lib/steep/cli.rb
ybiquitous: I defined `::Version`. Is this code correct?

lib/steep/cli.rb
```diff
@@ -20,6 +22,8 @@ def self.available_commands
     end
 
     def setup_global_options
+      OptionParser.new.parse(argv)
```
soutaro: I think we don't need to `parse` here. Right?

lib/steep/cli.rb
```diff
@@ -129,7 +133,7 @@ def process_interface
     end
 
     def process_version
-      stdout.puts Steep::VERSION
+      OptionParser.new.parse("--version")
```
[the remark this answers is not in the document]
soutaro: I don't think we have to fix this line.

lib/steep/cli.rb
ybiquitous: Do you think this way is correct?

```diff
     def setup_global_options
-      OptionParser.new.parse(argv)
+      version = OptionParser.new.version
+      if version
+        stdout.puts version
+        return false
+      end
 
       true
     end
```

This way outputs the only version without the program name (same as `steep version` subcommand).

```sh
# Before
$ ./exe/steep --version
steep 0.7.0

# After
$ ./exe/steep --version
0.7.0
```

ybiquitous: @soutaro ping?

soutaro: I know the difference, but don't think we should take care of that now. (I'll do that later.)

ybiquitous:
@soutaro I fixed all your suggestions. Please re-review when you have a time!

lib/steep/cli.rb
soutaro: I mean you don't have to do anything in `setup_global_options`. Just leave it empty.

ybiquitous: When I try to leave `setup_global_options` empty, then `--version` or `-v` options don't work. 😢 

```
$ ./exe/steep --version
Unknown command: --version
  available commands: check, validate, annotations, scaffold, interface, version
```

Here is my change:

```diff
     def setup_global_options
-      version = OptionParser.new.version
-      if version
-        stdout.puts version
-        return false
-      end
-
       true
     end
 ```

soutaro: Ah, got it.

soutaro:
@ybiquitous Squash the commits please.

ybiquitous:
I fixed up to one commit. 🙆‍♂️

soutaro:
@ybiquitous Merged. Thank you!

PR #58, opened 2018-10-09
Update built-in types: `Regexp`, `MatchData`

ybiquitous:
This PR updates built-in types `Regexp` and `MetaData`.

- `Regexp`:
  - Add missing methods.
  - API Doc: http://ruby-doc.org/core-2.5.1/Regexp.html
- `MatchData`:
  - Add class and methods.
  - API Doc: http://ruby-doc.org/core-2.5.1/MatchData.html
- `Encoding`:
  - `Regexp#encoding` requires `Encoding` type.
  - API Doc: http://ruby-doc.org/core-2.5.1/Encoding.html

### DIFF
--- a/smoke/regexp/a.rb
+++ b/smoke/regexp/a.rb
@@ -1,0 +1,147 @@
+# ALLOW FAILURE
+
+new_1 = Regexp.new("a")
+# !expects NoMethodError: type=::Regexp, method=foo
+new_1.foo
+
+new_2 = Regexp.new("a", true)
+# !expects NoMethodError: type=::Regexp, method=foo
+new_2.foo
+
+new_3 = Regexp.new("a", Regexp::EXTENDED | Regexp::IGNORECASE)
+# !expects NoMethodError: type=::Regexp, method=foo
+new_3.foo
+
+new_4 = Regexp.new(/a/)
+# !expects NoMethodError: type=::Regexp, method=foo
+new_4.foo
+
+compile_1 = Regexp.compile("a")
+# !expects NoMethodError: type=::Regexp, method=foo
+compile_1.foo
+
+compile_2 = Regexp.compile("a", true)
+# !expects NoMethodError: type=::Regexp, method=foo
+compile_2.foo
+
+compile_3 = Regexp.compile("a", Regexp::EXTENDED | Regexp::IGNORECASE)
+# !expects NoMethodError: type=::Regexp, method=foo
+compile_3.foo
+
+compile_4 = Regexp.compile(/a/)
+# !expects NoMethodError: type=::Regexp, method=foo
+compile_4.foo
+
+escape_1 = Regexp.escape("a")
+# !expects NoMethodError: type=::String, method=foo
+escape_1.foo
+
+last_match_1 = Regexp.last_match
+# !expects NoMethodError: type=(::MatchData | nil), method=foo
+last_match_1.foo
+
+last_match_2 = Regexp.last_match(1)
+# !expects NoMethodError: type=(::String | nil), method=foo
+last_match_2.foo
+
+quote_1 = Regexp.quote("a")
+# !expects NoMethodError: type=::String, method=foo
+quote_1.foo
+
+try_convert_1 = Regexp.try_convert(Object.new)
+# !expects NoMethodError: type=(::Regexp | nil), method=foo
+try_convert_1.foo
+
+union_1 = Regexp.union
+# !expects NoMethodError: type=::Regexp, method=foo
+union_1.foo
+
+union_2 = Regexp.union("a")
+# !expects NoMethodError: type=::Regexp, method=foo
+union_2.foo
+
+union_3 = Regexp.union("a", "b")
+# !expects NoMethodError: type=::Regexp, method=foo
+union_3.foo
+
+union_4 = Regexp.union(["a", "b"])
+# !expects NoMethodError: type=::Regexp, method=foo
+union_4.foo
+
+union_5 = Regexp.union(/a/)
+# !expects NoMethodError: type=::Regexp, method=foo
+union_5.foo
+
+union_6 = Regexp.union(/a/, /b/)
+# !expects NoMethodError: type=::Regexp, method=foo
+union_6.foo
+
+union_7 = Regexp.union([/a/, /b/])
+# !expects NoMethodError: type=::Regexp, method=foo
+union_7.foo
+
+op_eqeqeq_1 = /a/ === "a"
+# !expects NoMethodError: type=bool, method=foo
+op_eqeqeq_1.foo
+
+op_match_1 = /a/ =~ "a"
+# !expects NoMethodError: type=::Integer, method=foo
+op_match_1.foo
+
+casefold_1 = /a/.casefold?
+# !expects NoMethodError: type=bool, method=foo
+casefold_1.foo
+
+encoding_1 = /a/.encoding
+# !expects NoMethodError: type=::Encoding, method=foo
+encoding_1.foo
+
+fixed_encoding_1 = /a/.fixed_encoding?
+# !expects NoMethodError: type=bool, method=foo
+fixed_encoding_1.foo
+
+match_1 = /a/.match("a")
+# !expects NoMethodError: type=(::MatchData | nil), method=foo
+match_1.foo
+
+match_2 = /a/.match("a", 0)
+# !expects NoMethodError: type=(::MatchData | nil), method=foo
+match_2.foo
+
+/a/.match("a") do |m|
+  # !expects NoMethodError: type=::MatchData, method=foo
+  m.foo
+end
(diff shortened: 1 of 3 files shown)


stdlib/builtin.rbi
```diff
@@ -560,10 +560,60 @@ class Enumerator<'a, 'b>
...
+  def self.last_match: -> MatchData?
+                     | (Integer) -> String?
+  def self.quote: (String) -> String
+  def self.try_convert: (any) -> Regexp?
+  def self.union: (*String) -> Regexp
+                | (Array<String>) -> Regexp
+                | (*Regexp) -> Regexp
+                | (Array<Regexp>) -> Regexp
+
+  def initialize: (String, *any) -> any
```
ybiquitous: Perhaps we may be able to use more strict type instead of `*any`, but I don't know how to define such a type... :sweat:

```diff
-def initialize: (String, *any) -> any
+def initialize: (String) -> any
               | (String, Integer) -> any
               | (String, **non_nil_or_false**) -> any
```

API Doc says:

> If options is an Integer, it should be one or more of the constants Regexp::EXTENDED, Regexp::IGNORECASE, and Regexp::MULTILINE, or-ed together. Otherwise, if options is not nil or false, the regexp will be case insensitive.

http://ruby-doc.org/core-2.5.1/Regexp.html#method-c-new

(Also `Regexp.compile`)

soutaro: What we want exactly should be:

```
def initialize: (String) -> any           # no extra parameter
              | (String, Integer) -> any  # for Regexp options are given
              | (String, bool) -> any     # for any other values
```

stdlib/builtin.rbi
```diff
@@ -560,10 +560,60 @@ class Enumerator<'a, 'b>
...
+                  | (Regexp) -> any
+  def ===: (String) -> bool
+  def =~: (String) -> Integer?
+  def casefold?: -> bool
+  def encoding: -> Encoding
+  def fixed_encoding?: -> bool
+  def match: (String) -> MatchData?
+           | (String, Integer) -> MatchData?
+           | (String) { (MatchData) -> void } -> void
```
soutaro: `<'a> (String, Integer?) { (MatchData) -> 'a } -> ('a | nil)`

stdlib/builtin.rbi
```diff
@@ -560,10 +560,60 @@ class Enumerator<'a, 'b>
...
+  def options: -> Integer
   def source: -> String
+  def ~: -> Integer?
+end
+
+class MatchData
+  def []: (Integer | String | Symbol) -> String?
+        | (Integer, Integer) -> Array<String>
+        | (Range<Integer>) -> Array<String>
+  def begin: (Integer | String | Symbol) -> Integer
```
soutaro: `... -> Integer?`

stdlib/builtin.rbi
```diff
@@ -560,10 +560,60 @@ class Enumerator<'a, 'b>
...
+  def named_captures: -> Hash<String, String?>
+  def names: -> Array<String>
+  def offset: (Integer | String | Symbol) -> [Integer, Integer]
+  def post_match: -> String
+  def pre_match: -> String
+  def regexp: -> Regexp
+  def size: -> Integer
+  def string: -> String
+  def to_a: -> Array<String>
+  def values_at: (*(Integer | String | Symbol)) -> Array<String>
```
soutaro: `... -> Array<String?>`

stdlib/builtin.rbi
ybiquitous: I agree with your idea. 👍

stdlib/builtin.rbi
ybiquitous: Good! It works well. 😄 

I have one question. If changing the following,

```diff
   def match: (String) -> MatchData?
            | (String, Integer) -> MatchData?
-           | <'a> (String) { (MatchData) -> 'a } -> ('a | nil)
-           | <'a> (String, Integer) { (MatchData) -> 'a } -> ('a | nil)
+           | <'a> (String, Integer?) { (MatchData) -> 'a } -> ('a | nil)
```

some errors are raised. 😕 

```
$ bin/smoke_runner.rb smoke/regexp
🏇 Running smoke test in smoke/regexp...
  🤦‍♀️ Unexpected error found: smoke/regexp/b.rb:1:0: UnexpectedBlockGiven: method_type=(String) -> MatchData? (/(?<foo>a)/.match("a") do |match|)
  🤦‍♀️ Unexpected error found: smoke/regexp/a.rb:68:0: UnexpectedBlockGiven: method_type=(String) -> MatchData? (/a/.match("a") { |m| match_3 = m })
```

Can we make method override more simple?

stdlib/builtin.rbi
ybiquitous: Oh, this may be a ducumentation's bug. 💦 

Documentation says `begin(n) → integer`, but correctly `begin(n) → integer or nil`.

http://ruby-doc.org/core-2.5.1/MatchData.html#method-i-begin

Anyways, I will fix the `begin` definition. Thanks!

stdlib/builtin.rbi
ybiquitous: Similar problem as https://github.com/soutaro/steep/pull/58#discussion_r223859627. 😓 

When passing `Integer`, it can return `nil`.

```
$ irb
irb(main):001:0> match = /(?<foo>a)/.match("a")
=> #<MatchData "a" foo:"a">
irb(main):002:0> match.values_at(0,1,2,:foo)
=> ["a", "a", nil, "a"]
```

I will fix it. 👍

ybiquitous: I fixed via 8f4db9b, but actually should we do this? 👇 

```
def values_at: (*(String | Symbol)) -> Array<String>
             | (*(String | Symbol | Integer)) -> Array<String?>

**Comment by ybiquitous** [comment]
I fixed all issues, please review again! 🙏

File: stdlib/builtin.rbi [hunk 5]
```diff
@@ -564,7 +564,9 @@ class Encoding
 end
 
 class Regexp
-  def self.compile: (String, *any) -> Regexp
+  def self.compile: (String) -> Regexp
+                  | (String, bool) -> any
```
soutaro: The order of overloading matters.
`(String, bool) -> any` should be below (as far as I designed).

smoke/regexp/b.rb
```diff
@@ -10,7 +10,7 @@
   # @type var match_ref_5: Array<String>
   match_ref_5 = match[0..1]
 
-  # @type var begin_1: Integer
+  # @type var begin_1: Integer?
```
soutaro: To test the type of expression, the following would be correct and easy.

1. Assign to a variable without annotation
2. Call undefined method
3. Test the error message

soutaro: Example: https://github.com/soutaro/steep/blob/3a12024a089dab359ae05580adf2670dc3b5008c/smoke/type_case/a.rb#L13

stdlib/builtin.rbi
ybiquitous: Fixed via da3016a

smoke/regexp/b.rb
ybiquitous: As you pointed, I tested only normal cases, not abnormal cases. 😓 

When we test type definitions, the following points are necessary, right?

1. Valid argument(s) and return type of a method. For example:
    ```
    # @type var begin_1: Integer?
    begin_1 = match.begin(0)
    ```
2. Invalid argument(s) type of a method. For example:
    ```
    # !expects ArgumentTypeMismatch: receiver=::MatchData, expected=::Integer, actual=nil
    match.begin(nil)
    ```

soutaro:
You can use `$ steep check --dump-all-types` for debugging.

smoke/regexp/a.rb
```diff
@@ -0,0 +1,145 @@
...
+# !expects NoMethodError: type=::Regexp, method=foo
+union_7.foo
+
+op_eqeqeq_1 = /a/ === "a"
+# !expects NoMethodError: type=bool, method=foo
+op_eqeqeq_1.foo
+
+op_match_1 = /a/ =~ "a"
+# !expects NoMethodError: type=::Integer, method=foo
+op_match_1.foo
```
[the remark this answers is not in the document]
ybiquitous: Thanks! I'll try it later. 👍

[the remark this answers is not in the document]
ybiquitous: Done via 77b54ad

ybiquitous:
@soutaro Thanks a lot! I fixed all your suggestions! 😆

soutaro:
@ybiquitous Could you squash all the commits or fix them up?

ybiquitous:
I fixed up all commits to one commit. 🙆‍♂️

soutaro:
Thank you!

PR #64, opened 2018-10-28
Load types from gems

soutaro:
This PR is to allow shipping type definitions in gems. For this, gem developers and steep users should follow the protocol below.

## Gem developers

Put `*.rbi` files in a directory and ship them in the gem. And put a key/value pair `"steep_types" => signature_dir` in `metadata` of its gemspec.

```rb
spec.metadata = { "steep_types" => "sig" }
```

We recommend using `sig` as the name of the directory to have type definitions, but you can use any directory.

## Steep users

Run Steep with `-G` option, which accepts gem name `strong_json` or with version `strong_json:1.2.3`.

```
$ steep check -G strong_json -G other_gem -G another_gem lib
```

When you are using with bundler, `steep` detect that and load types from bundled gems automatically.

```
$ bundle exec steep check lib
```

`steep` also has `paths` subcommand, which print detected paths for signatures, and `--no-bundler` option, which disables bundled gem loading.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,5 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in steep.gemspec
 gemspec
+
+gem "with_steep_types", path: "test/gems/with_steep_types"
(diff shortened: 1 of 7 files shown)


ybiquitous:
Very nice feature!!! 😆 👍

PR #67, opened 2018-10-29
Unique signature paths

soutaro:
Fix type checking when developing a rubygem.

Because steep detects bundled gem signatures, the signatures of the developing gem loaded twice. Use `#uniq` to solve this issue.

### DIFF
--- a/lib/steep/cli.rb
+++ b/lib/steep/cli.rb
@@ -114,7 +114,7 @@ module Steep
           paths.unshift BUILTIN_PATH
         end
 
-        paths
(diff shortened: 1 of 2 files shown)


lib/steep/cli.rb
```diff
@@ -114,7 +114,7 @@ def paths
           paths.unshift BUILTIN_PATH
         end
 
-        paths
+        paths.reverse.uniq(&:realpath).reverse
```
soutaro: Reverse paths to let the `-I` path have higher priority.

PR #69, opened 2018-10-30
Fix Steep::Names::Base#hash

hanachin:
`@absolute` was renamed at 5000a506dd1be641819059db9ad0d18904a492f0

### DIFF
--- a/lib/steep/names.rb
+++ b/lib/steep/names.rb
@@ -24,7 +24,7 @@ module Steep
       end
 
       def hash
-        self.class.hash ^ name.hash ^ @absolute.hash
+        self.class.hash ^ name.hash ^ namespace.hash
       end
 
       alias eql? ==


soutaro:
Could you keep the `location` attribute?

I mean, `@absolute` instance variable is completely useless, which we should remove. But, I don't want to remove the `location` attribute.

hanachin:
@soutaro All right. I updated the pull request title and description.

lib/steep/names.rb
```diff
@@ -24,7 +24,7 @@ def ==(other)
       end
 
       def hash
-        self.class.hash ^ name.hash ^ @absolute.hash
+        self.class.hash ^ name.hash ^ @location.hash
```
soutaro: Could you drop `@location.hash`? Because `@location` is not used in `==`, so we should not use the value in `hash` too.

hanachin: Sorry for late response. I remove the `@location.hash`, 

Should we add `^ @namespace.hash`? `@namespace` looks used in `#==`
https://github.com/soutaro/steep/blob/v0.8.2/lib/steep/names.rb#L22-L24

soutaro: Yes. Please add `namespce.hash`. I really did not notice that...

hanachin: I updated the commit to compute the hash from the same values used for `#==`.
Thank you for everything!

soutaro:
Looks great! Thank you so much. 🎉

PR #71, opened 2018-10-31
Add `Kernel.#Integer` method's type definition

ybiquitous:
This PR adds the type definition of `Kernel.#Integer` method.
Please see API documentation: <http://ruby-doc.org/core-2.5.3/Kernel.html#method-i-Integer>

In addition, some types are added for smoke tests.

- `_ToI` (`interface`)
- `_ToInt` (`interface`) - see https://github.com/soutaro/steep/pull/71#discussion_r230157794
- ~~`Time` (implements `_ToI` interface)~~
  - Instead, add `WithToI` class for test
- `String#to_i` method
  - http://ruby-doc.org/core-2.5.3/String.html#method-i-to_i

Further, `Object#to_i` definition is removed, because `Object` doesn't have `to_i` method.
See also <http://ruby-doc.org/core-2.5.3/Object.html>.

### DIFF
--- a/smoke/integer/a.rb
+++ b/smoke/integer/a.rb
@@ -1,0 +1,31 @@
+integer_1 = Integer(1)
+# !expects NoMethodError: type=::Integer, method=foo
+integer_1.foo
+
+integer_2 = Integer("1")
+# !expects NoMethodError: type=::Integer, method=foo
+integer_2.foo
+
+class WithToInt
+  def to_int; 1; end
+end
+integer_3 = Integer(WithToInt.new)
+# !expects NoMethodError: type=::Integer, method=foo
+integer_3.foo
+
+class WithToI
+  def to_i; 1; end
+end
+integer_4 = Integer(WithToI.new)
+# !expects NoMethodError: type=::Integer, method=foo
+integer_4.foo
+
+integer_5 = Integer("10", 2)
+# !expects NoMethodError: type=::Integer, method=foo
+integer_5.foo
+
+# !expects IncompatibleArguments: receiver=::Object, method_type=(::String, ::Integer) -> ::Integer
+Integer(Object.new)
+
+# !expects IncompatibleArguments: receiver=::Object, method_type=(::String, ::Integer) -> ::Integer
+Integer(nil)
--- a/stdlib/builtin.rbi
+++ b/stdlib/builtin.rbi
@@ -12,7 +12,6 @@ class Object < BasicObject
   def ===: (any) -> bool
   def !=: (any) -> bool
   def class: -> class
-  def to_i: -> Integer
   def is_a?: (Module) -> bool
   def inspect: -> String
   def freeze: -> self
@@ -64,6 +63,8 @@ module Kernel
   def loop: { () -> void } -> void
   def puts: (*any) -> void
   def eval: (String, ? Integer?, ?String) -> any
+  def Integer: (String, Integer) -> Integer
+             | (_ToI | _ToInt) -> Integer
 end
 
 class Array<'a>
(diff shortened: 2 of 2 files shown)


stdlib/builtin.rbi
```diff
@@ -273,6 +278,7 @@ class Numeric
 end
 
 class Integer < Numeric
+  def to_i: -> Integer
```
ybiquitous: Removing `Object#to_i` causes the following error:

```
Unexpected error found: smoke/map/a.rb:5:15: NoMethodError: type=::Integer, method=to_i (x.to_i)
```

So, `Integer#to_i` must be added. `to_i` is an alias for `to_int`.
See <http://ruby-doc.org/core-2.5.3/Integer.html#method-i-to_i>.

smoke/integer/a.rb
```diff
@@ -0,0 +1,21 @@
...
+
+integer_4 = Integer("10", 2)
+# !expects NoMethodError: type=::Integer, method=foo
+integer_4.foo
+
+# !expects ArgumentTypeMismatch: receiver=::Object, expected=(::Numeric | ::String | ::_ToI), actual=::Object
+Integer(Object.new)
+
+# !expects ArgumentTypeMismatch: receiver=::Object, expected=(::Numeric | ::String | ::_ToI), actual=nil
+Integer(nil)
```
ybiquitous: I don't know that `smoke/integer/*.rb` is correct as smoke tests for `Kernle#.Integer`. 😓 

Please tell me if there are other good locations. 🙏

stdlib/builtin.rbi
```diff
@@ -64,6 +63,8 @@ module Kernel
   def loop: { () -> void } -> void
   def puts: (*any) -> void
   def eval: (String, ? Integer?, ?String) -> any
+  def Integer: (Numeric | String | _ToI) -> Integer
```
soutaro: It looks like `Integer` accepts any argument. So the following looks better to me.

```
def Integer: (String, Integer) -> Integer
           | (any) -> Integer
```

ybiquitous: Looks good! Surely, `String` has [`#to_i`](http://ruby-doc.org/core-2.5.3/String.html#method-i-to_i) method. 👍 

But I have a question. `Kernel.#Integer` documentation [says](http://ruby-doc.org/core-2.5.3/Kernel.html#method-i-Integer
):

> Non string values will be converted by first trying to_int, then to_i.

And `Numeric` has [`#to_int`](http://ruby-doc.org/core-2.5.3/Numeric.html#method-i-to_int) method but doesn't have `#to_i` method.

Then, should we have also `_ToInt` interface like the following? 👇 

```
def Integer: (String, Integer) -> Integer
           | (_ToInt | _ToI) -> Integer

interface _ToInt
  def to_int: -> Integer
end
```

stdlib/builtin.rbi
```diff
@@ -377,6 +387,8 @@ class String
   def empty?: -> bool
   def length: -> Integer
   def force_encoding: (any) -> self
+  def to_i: -> Integer
+          | (Integer) -> Integer
```
ybiquitous: `String` implements `_ToI` interface.
http://ruby-doc.org/core-2.5.3/String.html#method-i-to_i

stdlib/builtin.rbi
ybiquitous: I pushed 45fcacb and 8e61a5f. Please check it! 🙏

soutaro:
Looks good. Squash them please. 💪

ybiquitous:
Thanks. I squashed to one commit. 👌

soutaro:
Merged. Thanks. 🎉

PR #72, opened 2018-11-10
Support private

soutaro:
Private method can be declared as:

```
class Foo
  def (private) foo: () -> void
end
```

Only a method call without receiver can call the `foo` method.

```rb
class Foo
  def foo; end

  def bar
    foo()      # ok
    self.foo() # NoMethod error
  end
end
```

Steep validates type definition not to have a method definition override from public to private.

* You have to write `private` to all private methods
* If you don't write `private`, the method is *public*


soutaro:
I know the annotation syntax is weird, but I don't have any idea to improve that. We will fix that later.

PR #87, opened 2019-05-26
Relax rainbow dependency

meganemura:
It seems that the rainbow 3.0.0 is compatible with 2.2.2 for ruby >= 2.5.0 (steep required).
Changelog: https://github.com/sickill/rainbow/blob/master/Changelog.md#300-2017-11-29
https://rubygems.org/gems/rainbow


soutaro:
@meganemura Thanks! 🎉

Issue #93, opened 2019-09-12
Option or configuration to exclude paths


vinistock:
Hey! I am trying to adopt steep for a gem that includes generators. However, running `steep check` breaks on the generator's template files with the following error:

`#<Parser::SyntaxError: unterminated string meets end of file>`

Template files really don't need to be type checked, so I was wondering if there's a way to exclude certain paths from `steep check` either by a configuration file (say something like `.steep.yml` or by a command line argument).

I suspect the changes to exclude paths would have to be done somewhere in [here](https://github.com/soutaro/steep/blob/4bd719d6d92d41855a97dc77dfb77cfe0c6d4026/lib/steep/cli.rb#L146). Let me know your ideas and I can put a pull request together if you wish to implement this.

Thanks!

vinistock:
I see. Do you mind if I try to put together an implementation for a simple configuration file? I'd love to help.

soutaro:
@vinistock Sounds great.

I have written a document discussing the project file. My opinion is going though Ruby syntax, as in the *proposal* section. If you consider trying to implement, feedbacks are welcome!

https://hackmd.io/CrFK56eATgengAfZWyOI5Q

vinistock:
I really like the idea of the `Steepfile`. I am working on a sample implementation to get your ideas on it.

However, I am facing two issues (which might be related). When I try to run individual tests or open the console, I get the error `Cannot find ruby/signatures`. And if I run `bundle exec rake test` I end up in an endless loop. Am I missing something?

soutaro:
Ah, it is because `ruby-signature` is bundled as a git submodule now. Try `rake reset` to sync the submodule.

PR #94, opened 2019-09-20
Propose implementation of Steepfile DSL

vinistock:
Based on the discussion in #93. This PR is a simple proposal of a DSL for creating a Steepfile configuration file. The proposal is as follows:

1. Create a Dsl class that defines the available methods to be used in the Steepfile
2. Create a Configuration class to store merged options. These merged options will include Steep's defaults, Steepfile directives and command line arguments.

Let me know your feedback so that we can work towards the correct implementation.

Observations:

* The Configuration class is not yet used for the CLI class. If we decide that this implementation is the way to go, I can make changes to actually use it.
* At this moment, I only added the signatures directive to get a quick feedback on the implementation. We can incrementally move more of the options to the Dsl if so desired

### DIFF
--- a/lib/steep.rb
+++ b/lib/steep.rb
@@ -65,6 +65,8 @@ require "steep/type_inference/block_params"
 require "steep/type_inference/constant_env"
 require "steep/type_inference/type_env"
 require "steep/ast/types"
+require "steep/dsl"
+require "steep/configuration"
 
 require "steep/project"
 require "steep/project/file"
--- a/lib/steep/configuration.rb
+++ b/lib/steep/configuration.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+require "optparse"
+
+module Steep
+  class Configuration
+    class << self
+      # with_merged_options
+      #
+      # This method yields a new configuration object
+      # with merged results in the following precedence:
+      #
+      # 1. Command line arguments
+      # 2. Steepfile DSL arguments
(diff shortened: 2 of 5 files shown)


soutaro:
@vinistock I'm sorry for the late reply.

I'd ask to define a translation from `Steepfile` to `Project` object. I found there are some missing abstractions on `Project`, especially for LSP support, but I think working with `Project` is a good starting point.

I think to focus `check` subcommand is fine for now. I'd also suggest to skip merging CLI options and configuration file. The CLI files can be ignored if `Steepfile` exists.

So, the change should be:

1. Fix `drivers/check.rb` to read the configuration if `Steepfile` exists, and translate the `Steepfile` DSL to `Project`
    1. Translating the DSL directly to `Project` may work, but it would be better to define another intermediate structure (`Configuration`!)
2. If `Steepfile` doesn't exist, build the `Project` from the command line arguments

In my opinion, when the `Steepfile` implementation is completed, we can deprecated passing files directly from CLI.

soutaro:
@vinistock Hi, I had started working for project refactoring, and now I'm working for DSL implementation. https://github.com/soutaro/steep/tree/project

Thank you for taking your time and your efforts, but I would do it myself. 😃 Any feedback is welcome!

vinistock:
@soutaro sounds good! Sorry I couldn't get it done in time. I'll keep an eye on Steep and help in whatever I can!

PR #98, opened 2019-12-15
Do not make Steepfile in the project directory on test

pocke:
Currently `bundle exec rake` generates a `Steepfile` in the project directory unexpectedly.
This pull request will suppress it.

The test case makes a temporary directory, but actually it did not use the directory.
So this change makes it use the directory.


pocke:
My work looks conflicted with #99.

PR #132, opened 2020-04-26
Add type checking configuration to dsl

alts:
From the commit message:
> This allows the project owner to specify the strictness of type
> checks in their project. The naming of the new options draws upon
> the structure of similar options in mypy[0] and the naming of the
> typing errors themselves.
> 
> These two errors are very common ones that arise when trying to add
> steep to an existing untyped project.
> 
> [0] https://mypy.readthedocs.io/en/stable/config_file.html

I tried adding steep to a few projects just to see what the process of adding types to an existing project was. These two typing errors came up regularly, and could be an impediment to adding typing gradually to an existing project.

### DIFF
--- a/lib/steep/drivers/check.rb
+++ b/lib/steep/drivers/check.rb
@@ -56,9 +56,13 @@ module Steep
                   source_file.errors.reject do |error|
                     case
                     when error.is_a?(Errors::FallbackAny)
-                      !target.options.fallback_any_is_error
+                      target.options.allow_fallback_any
                     when error.is_a?(Errors::MethodDefinitionMissing)
                       target.options.allow_missing_definitions
+                    when error.is_a?(Errors::NoMethod)
+                      target.options.allow_unknown_method_calls
+                    when error.is_a?(Errors::UnknownConstantAssigned)
+                      target.options.allow_unknown_constant_assignment
                     end
                   end.each do |error|
                     error.print_to stdout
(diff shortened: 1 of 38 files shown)


soutaro:
@alts Thanks!

* Could you add a unit test for the new DSL syntax? I think adding a test case in `project_test.rb` makes sense.
* Could you check the test failures?

soutaro:
I took a look at the smoke test failures. The errors are because of that you implemented the method error filtering, which hasn't been implemented for months... 🙇‍♂️ 

I'd like to suggest to implement two major modes for type checking options, `strict` and `moderate` (or you may find better names.) In the smoke tests, we want to use the `strict` mode, like:

```rb
target ... do
  ...

  typing_options :strict
end
```

alts:
In that model, would `moderate` be the default settings?

alts:
The latest change introduces three strictness levels. `strict`, `default` (what existed before), and `lenient`. I can make `moderate` an alias for `default`, if you'd like. I also wonder about the long-term usefulness of `lenient` as a category rather than being accomplished through individual toggles.

I some assertions to `steepfile_test.rb`. I wasn't sure how to go about adding a useful test to `project_test.rb`. I also realize now that having the filtering of errors done in `Drivers::Check` probably isn't the best home. I tried moving it to `Project::Target`, which seems better for testability. I wasn't sure about how equivalent `TypeCheckStatus#type_check_sources` and `Target#source_files` were. It seems like there's some filtering occurring.

soutaro:
Thanks!

`:lenient` sounds good! And it seems that adding assertions on `steepfile_test` makes sense.

I agree that having the filtering in `Drivers::Check` doesn't look the best, but I'm also not very sure where is the best place... So, I think we can continue with your implementation.

PR #153, opened 2020-06-15
Allow rbs v0.4

pocke:
Thanks for releasing rbs v0.4.0 :tada:

But unfortunately, I couldn't use it because steep restricts rbs version. So this pull request relaxes the restriction to allow rbs v0.4.0.
I guess steep will work well with rbs v0.4.0 because `rake test` passes.


soutaro:
Thanks!

soutaro:
Released 0.17.1. https://rubygems.org/gems/steep/versions/0.17.1

PR #159, opened 2020-06-30
Sort methods in LanguageServer to return non-inherited methods first

zachhaitz:
* This ensures that methods defined on a class / module are returned
  earlier in the LSP method autocomplete
* The rationale being the assumption that the most used methods are the ones
  defined in the class rather than inherited ones.
* A _*future*_ improvment might make this sort by proximity to the current class,
  e.g. `class One < Two < Three` could return the methods defined in the source of `One`,
  then the methods defined in the source of `Two`, and then `Three`.

sort-methods

### DIFF
--- a/lib/steep/project/completion_provider.rb
+++ b/lib/steep/project/completion_provider.rb
@@ -10,7 +10,7 @@ module Steep
 
       InstanceVariableItem = Struct.new(:identifier, :range, :type, keyword_init: true)
       LocalVariableItem = Struct.new(:identifier, :range, :type, keyword_init: true)
-      MethodNameItem = Struct.new(:identifier, :range, :definition, :method_type, keyword_init: true)
+      MethodNameItem = Struct.new(:identifier, :range, :definition, :method_type, :inherited_method, keyword_init: true)
 
       attr_reader :source_text
       attr_reader :path
@@ -244,7 +244,8 @@ module Steep
                     items << MethodNameItem.new(identifier: name,
                                                 range: range,
                                                 definition: method,
-                                                method_type: method_type)
+                                                method_type: method_type,
+                                                inherited_method: inherited_method?(method, type))
                   end
                 end
               end
@@ -292,6 +293,10 @@ module Steep
         end
 
         index
+      end
+
+      def inherited_method?(method, type)
+        method.implemented_in&.name&.name != type.name&.name
       end
     end
   end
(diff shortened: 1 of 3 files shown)


lib/steep/server/interaction_worker.rb
```diff
@@ -160,7 +160,9 @@ def response_to_completion(path:, line:, column:, trigger:)
       def format_completion_item(item)
         range = LanguageServer::Protocol::Interface::Range.new(
           start: LanguageServer::Protocol::Interface::Position.new(
-            line: item.range.start.line-1,
+            line: item.range.start.line
```
soutaro, line 163: I think we need this `-1` because the line numbers in LSP is zero origin where its one origin in the rest of the world.

lib/steep/project/completion_provider.rb
```diff
@@ -293,6 +294,10 @@ def index_for(string, line:, column:)
 
         index
       end
+
+      def inherited_method?(method, type)
+        method.implemented_in.name&.name != type.name.name
```
soutaro, line 299: I think we should omit the last `.name`.

```suggestion
        method.implemented_in.name != type.name
```

lib/steep/server/interaction_worker.rb
zachhaitz: Oh! Totally didn't mean to remove that! Will add back

lib/steep/project/completion_provider.rb
zachhaitz: Ah, this doesn't seem to work right now.

It looks like `method.implemented_in.name.namespace` returns an `RBS::Namespace` and `type.name.namespace` returns a `Steep::AST::Namespace` so the equality operator in `Steep::Names::Base` fails since it relies on the namespaces being equal. 

We could change the equality operator for `Steep::AST::Namespace`? Not sure what relies on that equality or if the `CompletionProvider` is more special and it's the only place where we're comparing `RBS::Namespace`s to `Steep::AST::Namespace`.

soutaro: Ah, I got it. 💦

zachhaitz:
> Looks like it's good to merge.
> 
> @zachhaitz `push -f` if you want to fix up the commits. (I think I would squash the commits, but it's up to you.)

@soutaro I _think_ I need write access? but not sure!

soutaro:
@zachhaitz I could push to your repository! (I took care of that.)

PR #164, opened 2020-07-06
Don't autocomplete initialize

zachhaitz:
* initialize is a special method in ruby that developer
  wouldn't call on an instance of a class, so let's remove
  it for the autocomplete provided to LSP


zachhaitz:
thank you for the suggestion!

PR #169, opened 2020-07-21
Setup Stackprof

soutaro:


bin/steep-prof
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env ruby
+
+require "stackprof"
+
+mode = (ENV["STEEP_STACKPROF_MODE"] || :wall).to_sym
+out = ENV["STEEP_STACKPROF_OUT"] || "tmp/stackprof-#{mode}-steep.dump"
+
+def exit(*)
```
soutaro, line 8: Hijacking `exit` because calling it makes Stackprof stop working.

PR #173, opened 2020-07-29
Earlier shutdown with interruption while `steep watch`

tadd:
When I run `steep watch` by mistake, it took a long time to stop the steep process. This PR tries to make it shorter.

## Current Status of `master`

For example, if I run `steep watch` in a large project directory like [Mastodon](https://github.com/tootsuite/mastodon), it takes about **40 seconds** currently to stop the watching process with `Ctrl+C`:

```sh
$ time bundle exec steep --log-level=fatal watch app/models/
[deprecation warnings omitted]
👀 Watching directories, Ctrl-C to stop.
^CShutting down workers... # I typed Ctrl+C immediately after 👀 printed

real	0m40.593s
user	0m48.878s
sys	0m1.637s
```

## After PR Applied

It takes only **4 seconds** if this PR applied.

```sh
$ time bundle exec steep --log-level=error watch app/models/
[deprecation warnings omitted]
👀 Watching directories, Ctrl-C to stop.
^CShutting down workers...

real	0m4.045s
user	0m6.723s
sys	0m1.438s
```

### DIFF
--- a/lib/steep/server/base_worker.rb
+++ b/lib/steep/server/base_worker.rb
@@ -12,6 +12,7 @@ module Steep
         @project = project
         @reader = reader
         @writer = writer
+        @shutdown = false
       end
 
       def handle_request(request)
@@ -28,7 +29,7 @@ module Steep
           Steep.logger.formatter.push_tags(*tags)
           Steep.logger.tagged "background" do
             while job = queue.pop
-              handle_job(job)
+              handle_job(job) unless @shutdown
             end
           end
         end
@@ -38,11 +39,11 @@ module Steep
             reader.read do |request|
               case request[:method]
               when "shutdown"
-                # nop
+                @shutdown = true
               when "exit"
                 break
               else
-                handle_request(request)
+                handle_request(request) unless @shutdown
               end
             end
           ensure
--- a/lib/steep/server/master.rb
+++ b/lib/steep/server/master.rb
@@ -23,6 +23,7 @@ module Steep
         @signature_worker = signature_worker
         @code_workers = code_workers
         @worker_to_paths = {}
+        @shutdown = false
       end
 
       def start
@@ -59,7 +60,7 @@ module Steep
         end
 
         while job = queue.pop
-          writer.write(job)
+          writer.write(job) unless @shutdown
         end
 
         writer.io.close
@@ -154,6 +155,7 @@ module Steep
 
         when "shutdown"
           queue << { id: id, result: nil }
+          @shutdown = true
 
         when "exit"
           queue << nil


lib/steep/drivers/watch.rb
```diff
@@ -126,6 +126,10 @@ def run()
           client_writer.write({ method: :shutdown, id: 10000 })
           client_writer.write({ method: :exit })
           client_writer.io.close()
+          Thread.start do
+            sleep 1
+            master.interrupt if main_thread.alive?
```
tadd, line 131: Currently, if the running process was interrupted, `shutdown` and `exit` messages are sent.

I added to sleep for a while then interrupt the master worker because sending messages does not mean worker processes stops immediately.   `master.interrupt` yields all child worker processes are stopped soon.

lib/steep/server/worker_process.rb
```diff
@@ -57,6 +57,14 @@ def shutdown
...
         @wait_thread.join()
       end
+
+      def interrupt
+        return unless @wait_thread.alive?
+        Process.kill "INT", @wait_thread.pid
+        sleep 0.1
+        return unless @wait_thread.alive?
+        Process.kill "TERM", @wait_thread.pid
+      end
```
tadd, line 67: Worker processes can be stopped earlier by sending signals.

In my environment, SIGINT was not effective sometimes. Because of that, I added some lines to send a signal again with SIGTERM.

tadd:
Oops, I'll take care of the CI failure.

tadd:
After some discussion with Soutaro-san, we agreed to modify this PR.

1. Just request `shutdown` with the first `Interrupt` (Ctrl-C) (Just keep current behavior)
2. If the second Ctrl-C typed while waiting for termination of the main thread, wait for 5 seconds
3. If the main thread was not terminated, send `SIGTERM`to child processes and wait for 5 seconds again
4. If main thread still not terminated, send `SIGKILL` finally

So I modified and arranged my branch after rebasing on master. @soutaro could you confirm that the current code is what you intended?

lib/steep/drivers/watch.rb
```diff
@@ -129,7 +129,11 @@ def run()
         end
 
         listener.stop
-        main_thread.join
+        begin
+          main_thread.join
+        rescue Interrupt
+          master.interrupt unless main_thread.join(5)
```
tadd, line 135: After `Interrupt`, wait for `main_thread` again with timeout value.
If the timeout was occured, `master.intterupt` was called because  `Thread#join` returns `nil`.

tadd:
@soutaro Sorry for late. I rebased on master and force pushed with only first commit.

soutaro (approved):
Thanks! 💯

PR #200, opened 2020-09-03
Type checking fixes

soutaro:


lib/steep/server/code_worker.rb
```diff
@@ -139,6 +139,8 @@ def handle_request(request)
       end
 
       def handle_job(job)
+        sleep 0.1
```
soutaro, line 142: I'm not 100% sure if this works in other environments, but having a sleep here makes the frontend threads in code workers work before background threads.

PR #256, opened 2020-11-14
More flexible record typing

soutaro:
* Support alias to record types (bug fix!)
* Record types are now co-variant
* Allows skipping `nil`-able attributes

```
# @type var x: { name: String, email: String? }
x = { name: "foo", email: "foo@example.com" }          # OK
x = { name: "foo", email: nil }                        # OK
x = { name: "foo" }                                    # OK 🐕 
```

### DIFF
--- a/lib/steep/subtyping/check.rb
+++ b/lib/steep/subtyping/check.rb
@@ -342,23 +342,21 @@ module Steep
                   constraints: constraints)
 
           when relation.sub_type.is_a?(AST::Types::Record) && relation.super_type.is_a?(AST::Types::Record)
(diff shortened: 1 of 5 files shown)


ParadoxV5:
> * Allows skipping `nil`-able attributes
> 
> ```
> # @type var x: { name: String, email: String? }
> x = { name: "foo" }                                    # OK 🐕 
> ```

Ackchyually nonexistent key is distinguishable from `nil` values for non-implicit-nil methods such as `x.fetch :email`.

```ruby
# @type var x: { name: String, ?email: String? }
x = { name: "foo" }  # OK 💯
```

PR #259, opened 2020-11-15
Fix case-when typing (and && typing...)

soutaro:


test/type_construction_test.rb
```diff
@@ -6482,4 +6482,94 @@ def test_typing_record_nilable_attribute
       end
     end
   end
+
+  def test_type_case_case_when_assignment
```
soutaro, line 6486: Assignments in `when` branches without `else` were `nil`-able even if the case-when is exhaustive.

test/type_construction_test.rb
```diff
@@ -6482,4 +6482,94 @@ def test_typing_record_nilable_attribute
...
+      with_standard_construction(checker, source) do |construction, typing|
+        _, _, context = construction.synthesize(source.node)
+
+        assert_no_error typing
+        assert_equal parse_type("::String"), context.lvar_env[:a]
+      end
+    end
+  end
+
+  def test_type_case_case_selector
```
soutaro, line 6509: Variable type narrowing didn't work well if the condition of case-when is an assignment.

test/type_construction_test.rb
```diff
@@ -6482,4 +6482,94 @@ def test_typing_record_nilable_attribute
...
+      with_standard_construction(checker, source) do |construction, typing|
+        _, _, context = construction.synthesize(source.node)
+
+        assert_no_error typing
+        assert_equal parse_type("::String"), context.lvar_env[:a]
+      end
+    end
+  end
+
+  def test_bool_and_or
```
soutaro, line 6559: `&&` with logical types returned `bool || nil`.

PR #273, opened 2020-12-18
More colorful message

ybiquitous:
This change aims to make error messages on the terminal easier to view with more colors.
Note that this keeps the message format the same still.

## Before

image

## After

image

### DIFF
--- a/lib/steep/errors.rb
+++ b/lib/steep/errors.rb
@@ -8,7 +8,18 @@ module Steep
       end
 
       def location_to_str
-        Rainbow("#{node.loc.expression.source_buffer.name}:#{node.loc.first_line}:#{node.loc.column}").red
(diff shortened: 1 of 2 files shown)


lib/steep/errors.rb
```diff
@@ -127,7 +138,7 @@ def initialize(node:, type:, method:)
       end
 
       def to_s
-        "#{location_to_str}: NoMethodError: type=#{type}, method=#{method}"
+        format_message "type=#{type}, method=#{method}", class_name: "NoMethodError"
```
ybiquitous, line 141: [note] Only this class name is different from one in the message (`NoMethod` ≠ `NoMethodError`):

https://github.com/soutaro/steep/blob/6eec2fbd6ce85b498579f18e3df8aee171146a03/lib/steep/errors.rb#L119

soutaro (approved):
🌈

PR #292, opened 2021-01-01
Do not cast `untyped || something` to bool type

pocke:
# Problem

Steep casts `untyped || something` to bool type, but I think it is not the expected behavior.
For example:

```ruby
# test.rb

# @type var x: untyped
a = _ = nil
b = a || 42
b[0]
```

```ruby
# Steepfile
target :lib do
  check 'test.rb'
end
```

```console
$ steep check
test.rb:6:0: NoMethodError: type=bool, method=[] (b[0])
```

In this case, I expect `b` is also `untyped` because it is introduced by `a` (== `untyped`). But actually it is `bool`. So I got the `NoMethodError` unexpectedly.

# Solution

Stop casting `x || y` to `bool` type if the LHS (`x`) is `untyped`.

### DIFF
--- a/lib/steep/type_construction.rb
+++ b/lib/steep/type_construction.rb
@@ -1580,7 +1580,7 @@ module Steep
                   end
 
             type = case
-                   when check_relation(sub_type: left_type, super_type: AST::Builtin.bool_type).success?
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍

PR #303, opened 2021-02-06
Type checking with expected type errors

soutaro:
This PR is to add `--with-expectations` and `--save-expectations` option to `steep check`.

スクリーンショット 2021-02-07 0 52 36

* When `--with-expectations` option is specified, it confirms that if reported type errors are expected ones.
* When `--save-expectations` option is specified, it saves current type errors as _expectations file_.

The options are for running type checking in CI for the cases:

1. When type errors are expected (this would happen when you are writing RBS files)
2. When type errors cannot be removed (this would happen because of the poor experience provided by Steep 😢)

The workflow we assume is:

1. Run Steep and confirm if the current type errors are what you want.
2. Save the current status with `--save-expectations`.
3. Run Steep with `--with-expectations` in CI.

### ToDos

* [x] What if CLI arguments are given to `steep check`???
* [x] Update `test:output` tools.


palkan:
Hey @soutaro,

Just found this feature—really helpful!

What about making it possible to specify `expectations_path` in the `Steepfile` to always use it (without explicitly passing `--with-expectations`)?

PR #351, opened 2021-03-26
Optimize constant/global validation

soutaro:


lib/steep/services/signature_service.rb
```diff
@@ -318,13 +318,29 @@ def update_builder(ancestor_builder:, paths:)
       def type_names(paths:, env:)
         env.declarations.each.with_object(Set[]) do |decl, set|
           if decl.location
-            if paths.include?(decl.location.buffer.name)
+            if paths.include?(Pathname(decl.location.buffer.name))
```
soutaro, line 321: 🐛

lib/steep/services/type_check_service.rb
```diff
@@ -218,19 +218,44 @@ def validate_signature(path:, &block)
...
+                          validator.validate_one_alias(type_name)
+                        end
+                      end
+                    end
                   end
                 end
 
-                validator.validate_const()
-                validator.validate_global()
+                const_decls = service.const_decls(paths: Set[path], env: service.latest_env)
```
soutaro, line 238: Validate constants and globals only if the declaration is in `path`.

Issue #328, opened 2021-03-12
`steep check` raise UnresolvedOverloading for Array#to_h with block


ybiquitous:
## Description

`steep check` raises an `UnresolvedOverloading` error for the code using `Array#to_h` with a block.

Note: I'm not sure which this problem is caused by RBS or Steep, so I'll re-post this issue to [ruby/rbs](https://github.com/ruby/rbs) if the cause is RBS.

Here is the [latest RBS of `Array#to_h`](https://github.com/ruby/rbs/blob/e629aa7d9f5d55bcbba6632e598835a6651bc32c/core/array.rbs#L1855-L1856):

```ruby
class Array
  def to_h: () -> Hash[untyped, untyped]
          | [T, S] () { (Elem) -> [T, S] } -> Hash[T, S]
end
```

## Reproduction

<details open>
<summary>a.rb</summary>

```ruby
["abc", "d", ""].to_h { |e| [e, e.size] }
#=> {"abc"=>3, "d"=>1, ""=>0}
```

</details>

<details open>
<summary>Steepfile</summary>

```ruby
target :lib do
  check "*.rb"
end
```

</details>

<details open>
<summary>$ bundle exec steep check a.rb</summary>

```
# Type checking files:

F

a.rb:1:0: [error] Cannot find compatible overloading of method `to_h` of type `::Array[::String]`
│ Method types:
│   def to_h: () -> ::Hash[untyped, untyped]
│           | [T, S] () { (::String) -> [T, S] } -> ::Hash[T, S]
│
│ Diagnostic ID: Ruby::UnresolvedOverloading
│
└ ["abc", "d", ""].to_h { |e| [e, e.size] }
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Detected 1 problem from 1 file
```

</details>

## Environment

- Ruby: `ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin19]`
- Steep: `0.42.0`
- RBS: `rbs 1.1.0`

soutaro:
Thank you for reporting this. This should be a bug. You can workaround if you add a type annotation in the block:

```
["abc", "d", ""].to_h { |e|
  # @type block: [String, Integer]
  [e, e.size]
}

```

ybiquitous:
@soutaro Thanks for your feedback! The workaround works on my code. 👍

PR #359, opened 2021-03-27
More tuple types through hint

soutaro:
Checking subtype relation when typing tuple results in `Array` types when the tuple members are type variables as reported in #328. This PR is to skip the type checking.

Fixes https://github.com/soutaro/steep/issues/328


lib/steep/type_construction.rb
```diff
@@ -1644,8 +1644,7 @@ def synthesize(node, hint: nil, condition: false)
               if hint && !(tuples = select_flatten_types(hint) {|type| type.is_a?(AST::Types::Tuple) }).empty?
                 tuples.each do |tuple|
                   typing.new_child(node_range) do |child_typing|
-                    pair = with_new_typing(child_typing).try_tuple_type(node, tuple)
-                    if pair && pair.constr.check_relation(sub_type: pair.type, super_type: hint).success?
```
soutaro, line 1648: I don't remember but it seems like the `check_relation` is to find the best tuple types from union types.

```ruby
# @type var tuple: [Integer, String] | [1, "foo"]
tuple = [1, ""]
```

The implementation doesn't work as expected because it just returns the first matching type, not the best one. The first-check-first-return strategy doesn't make any sense because the order in union types is not preserved.

PR #360, opened 2021-03-27
Pretty table printing

soutaro:
`steep stats` now support _table_ and _csv_ printing modes. Users can specify the mode through `--format` options.

```shell
$ steep stats --format=table
 Target  File                                         Status   Typed calls  Untyped calls  All calls  Typed %
--------------------------------------------------------------------------------------------------------------
 test    argument_type_mismatch.rb                    success            0              0          1       0%
 test    block_body_type_mismatch.rb                  success            1              0          2      50%
 test    block_type_mismatch.rb                       success            0              2          3       0%
 test    break_type_mismatch.rb                       success            1              0          1     100%
```

The default mode depends on `STDOUT.tty?`. _table_ mode is used if it's `true` and _csv_ mode is used if it's `false`. This assumes a workflow to run `steep stats` command behind a pipe and save the result as a CSV file for further visualization/review/processing. This PR won't break the workflow.

### DIFF
--- a/lib/steep/cli.rb
+++ b/lib/steep/cli.rb
@@ -113,14 +113,7 @@ module Steep
           opts.banner = "Usage: steep stats [options] [sources]"
 
           opts.on("--steepfile=PATH") {|path| check.steepfile = Pathname(path) }
-          opts.on("--format=FORMAT", "Specify output format: csv, table") do |format|
-            case format
-            when "csv", "table"
-              check.format = format
-            else
-              stderr.puts "--format option is ignored because it has unknown value: `#{format}`"
-            end
-          end
+          opts.on("--format=FORMAT", "Specify output format: csv, table") {|format| check.format = format }
           handle_jobs_option check, opts
           handle_logging_options opts
         end.parse!(argv)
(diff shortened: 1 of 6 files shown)


lib/steep/cli.rb
```diff
@@ -113,6 +113,14 @@ def process_stats
           opts.banner = "Usage: steep stats [options] [sources]"
 
           opts.on("--steepfile=PATH") {|path| check.steepfile = Pathname(path) }
+          opts.on("--format=FORMAT", "Specify output format: csv, table") do |format|
+            case format
+            when "csv", "table"
+              check.format = format
+            else
+              stderr.puts "--format option is ignored because it has unknown value: `#{format}`"
```
ybiquitous, line 121: [IMO] Since `ArgumentError` occurs on an invalid `--log-level` value, it seems kind to me to raise `ArgumentError` also on an invalid `--format` value.

Error:

```console
$ bundle exec steep stats --log-level=fata
#<ArgumentError: invalid log level: data>
```

No error:

```console
$ bundle exec steep stats --log-level=fatal --format=tabl
--format option is ignored because it has unknown value: `tabl`
# Calculating stats:

 Target  File                                             Status   Typed calls  Untyped calls  All calls  Typed %
------------------------------------------------------------------------------------------------------------------
...
```

soutaro: Good point! Will do. 💪

Issue #338, opened 2021-03-16
`@implements` in blocks


soutaro:
Some methods including `#instance_eval` change the value of `self` inside the given block. `@type self` annotation is for this case, but it doesn't work for method definitions. We typically see this when we use `Struct` class.

```rb
ReadMessageJob = _ = Struct.new(:source, :message) do
  # @implements ReadMessageJob

  def request?
    message[:method] && message[:id]
  end
end
```

The problem is

1. `@implements` annotations inside blocks are ignored.
2. `@type self: T` and `@type module: T` are not for method definitions.

So, I'd like to extend `@implements` annotations to be placed inside blocks.

soutaro:
Doesn't work for:

```ruby
class Methods
  Definition = _ = Struct.new(:name, :type, :originals, :overloads, :accessibilities, keyword_init: true) do
    # @implements Definition

    def self.empty(name:, type:)
      new(type: type, name: name, originals: [], overloads: [], accessibilities: [])
    end
  end
end
```

The `new` call in `self.empty` is resolved to `Methods.new`. 💣

PR #366, opened 2021-03-29
Fix annotation calculation after `delete_unrelated_defs`

soutaro:
This breaks completion/hover in blocks #338

### DIFF
--- a/lib/steep/services/hover_content.rb
+++ b/lib/steep/services/hover_content.rb
@@ -43,6 +43,7 @@ module Steep
       def typecheck(target, path:, content:, line:, column:)
         subtyping = service.signature_services[target.name].current_subtyping or return
         source = Source.parse(content, path: path, factory: subtyping.factory)
+        source = source.without_unrelated_defs(line: line, column: column)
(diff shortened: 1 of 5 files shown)


lib/steep/services/hover_content.rb
```diff
@@ -43,6 +43,7 @@ def method_definition_for(factory, type_name, singleton_method: nil, instance_me
       def typecheck(target, path:, content:, line:, column:)
         subtyping = service.signature_services[target.name].current_subtyping or return
         source = Source.parse(content, path: path, factory: subtyping.factory)
+        source = source.without_unrelated_defs(line: line, column: column)
```
soutaro, line 46: This is for performance improvement.

lib/steep/source.rb
```diff
@@ -72,13 +72,13 @@ def self.parse(source_code, path:, factory:)
         end
       end
 
-      mapping = {}
+      mapping = {}.compare_by_identity
```
soutaro, line 75: I know what I want is `#compare_by_identity` today. 🆔

PR #370, opened 2021-04-14
Do not crash when completion request context is missing

OlegSmelov:
Sublime LSP plugin doesn't send the `context` when doing a `textDocument/completion` request:

```
:: --> ruby textDocument/completion(7): {'position': {'line': 10, 'character': 21}, 'textDocument': {'uri': 'file://[redacted]/app.rb'}}
```

This causes Steep to crash with `NoMethodError (undefined method '[]' for nil:NilClass)`.

I'm not sure if this is a violation of the protocol, but it looks like `trigger` isn't used for anything important in Steep, so let's handle it gracefully.


soutaro (approved):
Thanks! The change totally makes sense.

* `context` is an optional parameter, and Steep should handle the case.
* `trigger` is not used at all.

Issue #336, opened 2021-03-16
Support Ruby 3 syntax


soutaro:
New syntaxes including:

* Numbered parameters
* `case-in`
* anything else?

akimd:
steep does choke on endless defs.

tekwizz123-mobile:
Is this resolved now that the linked PR has been closed? Or are there still remaining issues to be resolved? Just trying to evaluate if this is something that could be worked on.

PR #381, opened 2021-05-09
Ruby3 support

soutaro:
Working for #336.

This PR implements support for end-less defs and numbered parameters.


test/test_helper.rb
```diff
@@ -5,8 +5,6 @@
 require "bundler/setup"
 require 'steep'
 
-require "minitest/reporters"
-Minitest::Reporters.use! [Minitest::Reporters::DefaultReporter.new]
```
soutaro, line 9: RubyMine complained about the use of `Minitest::Reporters`.

PR #385, opened 2021-06-11
Display type information when type alias is hovered in rbs

Yuuki77:
we now display type information when alias is hovered in rbs 
image

## Known issue
The judgment when hovering is off by one character.
I think this is a location issue, so I'll fix it in a later PR.

### DIFF
--- a/lib/steep/server/interaction_worker.rb
+++ b/lib/steep/server/interaction_worker.rb
@@ -80,9 +80,16 @@ module Steep
             content = hover.content_for(path: job.path, line: job.line, column: job.column+1)
             if content
               range = content.location.yield_self do |location|
-                start_position = { line: location.line - 1, character: location.column }
-                end_position = { line: location.last_line - 1, character: location.last_column }
-                { start: start_position, end: end_position }
+                case content.location
+                when RBS::Location::WithChildren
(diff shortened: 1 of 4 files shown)


lib/steep/server/interaction_worker.rb
```diff
@@ -80,9 +80,16 @@ def process_hover(job)
             content = hover.content_for(path: job.path, line: job.line, column: job.column+1)
             if content
               range = content.location.yield_self do |location|
-                start_position = { line: location.line - 1, character: location.column }
-                end_position = { line: location.last_line - 1, character: location.last_column }
-                { start: start_position, end: end_position }
+                case content.location
```
soutaro, line 83: You can use `#as_lsp_range` (https://github.com/soutaro/steep/blob/master/lib/steep/range_extension.rb)

Yuuki77: thank you for the review!
updated it here https://github.com/soutaro/steep/pull/385/commits/d93aee587ef7710723f6f5791b8af0989d8682b7

soutaro (approved):
👍

soutaro:
@Yuuki77 One of the tests is failing...

Yuuki77:
@soutaro sorry I just fixed the test
it should be ok now 💪

PR #396, opened 2021-07-09
WIP: Object in main interface send

### LINKED ISSUE #392: Object methods can't be found for interface
The following signature and code:

```ruby
# sig
def a: () -> _ToS

# ruby 
a == "A"
```

throw the error:

```
Type `::_ToS` does not have method `==`(Ruby::NoMethod)
```

This happens because `Object` rbs methods aren't taken into account for interface return values. I think that they should.

HoneyryderChuck:
This PR integrates `Object` methods on method call checks.

The purpose is for calls to "implicit methods" such as `.nil?`, `==` or `.to_s` to not be tagged by `steep` as not implemented.

Fixes #392

### DIFF
--- a/lib/steep/ast/types/intersection.rb
+++ b/lib/steep/ast/types/intersection.rb
@@ -43,7 +43,8 @@ module Steep
         end
 
         def ==(other)
-          other.is_a?(Intersection) && other.types == types
+          (other.is_a?(Intersection) && other.types == types) ||
+          types.any? { |typ| typ == other }
         end
 
         def hash
--- a/lib/steep/ast/types/union.rb
+++ b/lib/steep/ast/types/union.rb
@@ -45,7 +45,9 @@ module Steep
 
         def ==(other)
           other.is_a?(Union) &&
-            Set.new(other.types) == Set.new(types)
+           other.types.all? do |typ1|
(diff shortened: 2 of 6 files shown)


lib/steep/ast/types/intersection.rb
```diff
@@ -43,7 +43,8 @@ def self.build(types:, location: nil)
         end
 
         def ==(other)
-          other.is_a?(Intersection) && other.types == types
+          (other.is_a?(Intersection) && other.types == types) ||
```
soutaro, line 46: The `==` on types are to test the equality of the values. You should use `Subtyping::Check` to test semantic relation between types.

soutaro:
I'm good to add an option to add methods of `Object` or `BasicObject` to interface types.
Seems like fixing `Steep::AST::Types::Factory` would be the best way to implement it.

HoneyryderChuck:
> Seems like fixing Steep::AST::Types::Factory would be the best way to implement it.

I think I did this at some points, but things started faialinng somewhere else.

TBH I'm not sure if I can finish this. I can't entirely follow how the analyzer traverses the code, so I'm a bit unsure if I can do this myself :/

PR #397, opened 2021-07-09
Extend the functionality of hover in rbs files.

Yuuki77:
add hover supports for `RBS::Types::ClassInstance`, `RBS::Types::ClassSingleton` and `RBS::Types::Interface`.

When these are hovered, the information and comments will now be displayed!

image

### How to test
crate a rbs file with this content and you can confirm that you can see the information when you hover them
```rbs
# my comment
class Foo [T] < Parent[T]
end

class Parent [in T]
end

module Hoge
end

class Qux
  @foo: Foo[Hoge]
end

interface _Fooable
  def foo: () -> nil
end

class C
  def foo: (_Fooable) -> singleton(String)
end
```

### DIFF
--- a/lib/steep/services/hover_content.rb
+++ b/lib/steep/services/hover_content.rb
@@ -6,6 +6,9 @@ module Steep
       MethodCallContent = Struct.new(:node, :method_name, :type, :definition, :location, keyword_init: true)
       DefinitionContent = Struct.new(:node, :method_name, :method_type, :definition, :location, keyword_init: true) do
       TypeAliasContent = Struct.new(:location, :decl, keyword_init: true)
+      ClassContent = Struct.new(:location, :decl, keyword_init: true)
+      InterfaceContent = Struct.new(:location, :decl, keyword_init: true)
+
         def comment_string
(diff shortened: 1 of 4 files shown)


lib/steep/services/hover_content.rb
```diff
@@ -76,6 +79,25 @@ def content_for(path:, line:, column:)
...
+                decl: class_decl
+              )
+            end
+          when RBS::Types::Interface
+            env = service.latest_env
+            interface_decl = env.interface_decls[type.name]&.decl or raise
+            location = type.location
+
+            InterfaceContent.new(
+              location: location,
```
soutaro, line 98: Use `location[:name]` here to highlight only name of the interface.

soutaro (changes requested):
The rest of the code looks good.

Yuuki77:
@soutaro Thank you for the reviews!
I have updated it https://github.com/soutaro/steep/pull/397/commits/1545d6b6b34a85bde47f2c253673c359dacec421

soutaro (approved):
💪

PR #403, opened 2021-08-10
Should be ignored hover for none target files

ksss:
I use steep on vscode with master version ( 43f896d ).

But it rails an error when hover with none target files.

image

I fixed it.


ksss:
It seems resolved from https://github.com/soutaro/steep/pull/404
Thanks!

PR #405, opened 2021-08-10
Fix NoMethodError when csend

ksss:
I encountered the following error.

image

And I found out that the backtrace is happening for `csend` at the fix point.
I hope this error will fix.


ksss:
Changed the solution.
ref https://github.com/soutaro/steep/pull/408

PR #411, opened 2021-08-20
Fix Unexpected error

ksss:
If `RBS::MixinClassError` is raised during a steep check, the steep worker thread will be in a wait state forever.

With this fix, not only will steep not stop when `RBS::MixinClassError` is raised, but it will also show the problem.

```
mixin-class-error.rbs:5:2: [error] Cannot include a class `::Foo` in the definition of `::Bar`
│ Diagnostic ID: RBS::MixinClassError
│
└   include Foo
    ~~~~~~~~~~~
```

### DIFF
--- a/lib/steep/diagnostic/signature.rb
+++ b/lib/steep/diagnostic/signature.rb
@@ -275,6 +275,34 @@ module Steep
         end
       end
 
+      class MixinClassError < Base
+        attr_reader :member
+        attr_reader :type_name
+
+        def initialize(location:, member:, type_name:)
+          super(location: location)
+          @member = member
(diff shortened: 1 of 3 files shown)


soutaro (approved):
Thank you! 🎉 🖊️

soutaro:
Oops the implementation looks good, but still has test failure...

soutaro:
@ksss The test failures should be fixed in #413. Could you try rebasing on the `master` branch?

soutaro:
I run test locally and confirmed it passes.

ksss:
Thanks ✨

PR #412, opened 2021-08-22
Fix NoMethodError with csend

soutaro:
Thanks to @ksss for reporting the issues. 🙏 

* Closes #405
* Closes #408

Seems like a bug in Diagnostic classes.


ksss:
GREAT ❤️

PR #420, opened 2021-08-28
Enable RBS collection feature

pocke:
This PR introduces `rbs collection` feature, https://github.com/ruby/rbs/pull/589,  to Steep.

# Usage

* If `rbs_collection.yaml` exists in the same directory as `Steepfile`, Steep loads libraries' RBSs from the config file.
* If `collection_config path` is specified in Steepfile, Steep loads libraries' RBSs from the specified file.
  * I suppose the configuration will be used like `collection_config '../other/path/rbs_collection.yaml'` in many cases.
* If `disable_collection` is specified in Steepfile, Steep ignores `rbs_collection.yaml`.
* If `rbs_collection.yaml` doesn't exist and no options are specified, Steep does nothing about rbs collection feature.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ PATH
       parallel (>= 1.0.0)
       parser (>= 3.0)
       rainbow (>= 2.2.2, < 4.0)
-      rbs (>= 1.2.0)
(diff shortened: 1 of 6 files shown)


soutaro (approved):
👏 🍔

PR #426, opened 2021-09-01
Fix error when hover on comment for rbs

ksss:
Fixed an error that occurred when hovering the IDE cursor over a comment in rbs file.


soutaro (approved):
🙏

PR #438, opened 2021-09-20
Fix NoMethodError on block params with rest var

ksss:
Steep will raise UnexpectedError when non-annotated block params with rest variable.

```rb
a = ->(*res) { }
# => UnexpectedError: undefined method `args' for nil:NilClass
```


soutaro (approved):
Thanks! 👍

PR #442, opened 2021-10-01
Adjusting display of CompletionItem

ksss:
There are two problems with the current completion function.

- If I want to complete a method such as `default_value`, `def` part will be completed.
- Type parameters are also being completed.

https://user-images.githubusercontent.com/935310/135632516-706616e6-c46a-420e-b804-008274931427.mov

So, the following fixes have been made.

- Remove the `def ` part from the completion target.
- Remove type parameters from the completion target as well. Instead, put type parameters in `detail`.

This solves the problem, but the downside is that the type parameter that used to show up in all completion lists is now only the one you have selected.

https://user-images.githubusercontent.com/935310/135633224-634bcbf4-f61f-4502-bb24-4607656aea62.mov

### DIFF
--- a/lib/steep/server/interaction_worker.rb
+++ b/lib/steep/server/interaction_worker.rb
@@ -383,7 +383,7 @@ HOVER
         when RBS::AST::Declarations::Interface
           "interface #{name_and_params(decl.name, decl.type_params)}"
         end
-    end
+      end
 
       def format_completion_item(item)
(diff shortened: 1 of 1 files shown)


soutaro (approved):
Looks great! Thank you for sending the patch!

(I was thinking of using `filterText` attribute in fact, but deleting `def` and the types would look better.)

Issue #429, opened 2021-09-05
Ruby::UnexpectedError when a method receives a kwarg as a record type


pocke:
`Ruby::UnexpectedError` occurs when a method receives a kwarg as a record type.

```ruby
# Steepfile
target :lib do
  check '.'
  signature '.'
end
```

```ruby
# test.rbs
class C
  def self.m1: ({a: String}) -> void
  def self.m2: (a: String) -> void
  def self.m3: (Hash[Symbol, String]) -> void
end
```

```ruby
# test.rb
C.m1(a: 'str')     # error
C.m1({a: 'str'})   # ok
C.m2(a: 'str')     # ok
# C.m2({a: 'str'}) # Type error (expected)
C.m3(a: 'str')     # ok
C.m3({a: 'str'})   # ok
```

```console
$ steep check
```

ksss:
Maybe resolved.
https://github.com/soutaro/steep/pull/444

PR #444, opened 2021-10-13
Fix Unexpected error if record-type receive kwargs

ksss:
https://github.com/ruby/gem_rbs_collection/pull/68 seems raise Unexpected error.
I've investigated and fixed it, because it seems that `Steep::TypeConstruction#type_hash_record` didn't consider the case of kwargs enough.


ksss:
Maybe it's related to https://github.com/soutaro/steep/issues/429 .
@pocke What do you think?

soutaro (approved):
🙏

PR #448, opened 2021-11-09
Add a missing attr_reader in the example code

kymmt90:
The Phone class should have `attr_reader  :country, :number` if Phone#== uses `other.country` and `other.number`.


soutaro (approved):
Nice catch! 👏

PR #449, opened 2021-11-11
Should check if positional_params is nil or not

ksss:
The following code raises an unexpected error.

```rb
[].each(*a)
```

```
Unexpected error: #<NoMethodError: undefined method `each' for nil:NilClass>
/Users/ksss/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/steep-0.46.0/lib/steep/type_inference/send_args.rb:156:in `uniform_type'
/Users/ksss/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/steep-0.46.0/lib/steep/type_inference/send_args.rb:567:in `block in each'
...
```

### DIFF
--- a/lib/steep/type_inference/send_args.rb
+++ b/lib/steep/type_inference/send_args.rb
@@ -153,6 +153,7 @@ module Steep
         end
 
         def uniform_type
+          return nil unless positional_params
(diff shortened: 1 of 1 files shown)


soutaro (approved):
👍

PR #495, opened 2022-02-24
RBS 2.2

soutaro:
This PR makes Steep run with RBS 2.2, without supporting the new features.

* Bounded generics are ignored (the constraint is ignored)
* Generic type alias is ignored (all type parameters are treated as `untyped`)

I'm working to support these type checking features, but you can use Steep with other RBS 2.2 features.


soutaro:
This includes some commits from #479. Thank you @takahashim. 🙏 

Closes #479

Issue #515, opened 2022-03-18
Unexpected error when or-asgn/and-asgn


ksss:
issue.rbs

```rbs
class Issue
  def unexpected=: (Integer) -> Integer
end
```

issue.rb

```rb
class Issue
  def or_asgn
    self.unexpected ||= 1
  end
end
```

Steepfile

```rb
target :issue do
  check "issue.rb"
  signature "issue.rbs"

  configure_code_diagnostics(Steep::Diagnostic::Ruby.all_error)
end
```

`$ steep check`

```
[Steep 0.49.1] [typecheck:typecheck@0] [background] [#typecheck_source(path=sample.rb)] [#type_check_file(sample.rb@sample)] [synthesize:(1:1)] [synthesize:(70:1)] [synthesize:(71:3)] [synthesize:(72:5)] [synthesize:(72:5)] [synthesize:(72:5)] Unexpected error: #<NoMethodError: undefined method `type' for nil:NilClass

                 arguments.take_while {|node| node.type != :block_pass }
                                                  ^^^^^>
[Steep 0.49.1] [typecheck:typecheck@0] [background] [#typecheck_source(path=sample.rb)] [#type_check_file(sample.rb@sample)] [synthesize:(1:1)] [synthesize:(70:1)] [synthesize:(71:3)] [synthesize:(72:5)] [synthesize:(72:5)] [synthesize:(72:5)]   /Users/ksss/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/steep-0.49.1/lib/steep/type_inference/send_args.rb:507:in `block in positional_arg'
[Steep 0.49.1] [typecheck:typecheck@0] [background] [#typecheck_source(path=sample.rb)] [#type_check_file(sample.rb@sample)] [synthesize:(1:1)] [synthesize:(70:1)] [synthesize:(71:3)] [synthesize:(72:5)] [synthesize:(72:5)] [synthesize:(72:5)]   /Users/ksss/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/steep-0.49.1/lib/steep/type_inference/send_args.rb:507:in `take_while'
[...]

PR #517, opened 2022-03-20
Fix unexpected error when or-asgn/and-asgn

ksss:
Fix https://github.com/soutaro/steep/issues/515

or-asgn/and-asgn may differ in the number of children under the send node.

```
$ ruby-parse --legacy -e 'a=[]; a[0] ||= 3'
(begin
  (lvasgn :a
    (array))
  (or-asgn
    (send
      (lvar :a) :[]
      (int 0)) # <= `(int 0)` is asgn.children[2]
    (int 3)))

$ ruby-parse --legacy -e 'self.a ||= 3'
(or-asgn
  (send
    (self) :a) # <= not have asgn.children[2]
  (int 3))
```

This PR support both pattern.

### DIFF
--- a/lib/steep/type_construction.rb
+++ b/lib/steep/type_construction.rb
@@ -2275,13 +2275,10 @@ module Steep
               type, constr = synthesize(rhs, hint: hint)
               constr.ivasgn(asgn, type)
             when :send
-              rhs_ = node.updated(:send,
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Nice catch! 👍 👏

PR #561, opened 2022-05-28
Fix: Fix steep not working on windows

sevenc-nanashi:
This PR will fix steep not working on windows.
Note the test for windows is not done yet. (It deadlocks somehow)

### DIFF
--- a/.github/workflows/ruby-windows.yml
+++ b/.github/workflows/ruby-windows.yml
@@ -1,4 +1,4 @@
-name: Ruby
+name: Run test on Windows
 
 on:
   push:
--- a/.github/workflows/ruby-windows.yml
+++ b/.github/workflows/ruby-windows.yml
@@ -16,7 +16,7 @@ jobs:
         - "3.1"
         task:
         - test
-        - test:output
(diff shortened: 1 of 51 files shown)


.github/workflows/ruby-windows.yml
```diff
@@ -0,0 +1,34 @@
+name: Run test on Windows
```
soutaro, line 1: 👏 👏 👏

lib/steep/project.rb
```diff
@@ -16,22 +16,10 @@ def base_dir
...
-               end
-               Pathname.new(
-                 path_str
-               )
-             else
-               orig_path
-             end
+    def relative_path(path)
       path.relative_path_from(base_dir)
+    rescue ArgumentError
```
soutaro, line 21: Could you tell me what is raising `ArgumentError` here?

sevenc-nanashi: Pathname#relative_path_from raises ArgumentError when Drive letter is different.

soutaro: Got it. It's really interesting.
Let's think about it in the future, when problem happens.

Issue #526, opened 2022-04-03
`steep langserver` stacks when unsaved ruby file is opened in VSCode


autopp:
In VSCode, when we open a new file and set Ruby to its syntax, `langserver` stacks.

### Procedure for reproducing

- Add `steep` to `Gemfile`
- Create and setup `Steepfile`
- Create a new file (`Untitled-1` in editor) in VSCode
- Set `Ruby` to the file by `Change Language Mode`
- Do `Steep: Restart all`

<img width="469" alt="image" src="https://user-images.githubusercontent.com/7571111/161430066-54c7223a-385f-44ae-848f-dd400d942f41.png">

<details>
<summary>Output :</summary>

```
[Steep 0.51.0] [master] [main] [ReceiveMessageJob(client/textDocument/didOpen/)] Processing message from client: method=textDocument/didOpen, id=
#<TypeError: no implicit conversion of nil into String>
  /Users/autopp/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/steep-0.51.0/lib/steep/server/master.rb:466:in `initialize'
  /Users/autopp/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/steep-0.51.0/lib/steep/server/master.rb:466:in `Pathname'
  /Users/autopp/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/steep-0.51.0/lib/steep/server/master.rb:466:in `pathname'
  /Users/autopp/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/steep-0.51.0/lib/steep/server/master.rb:543:in `process_message_from_client'
  /Users/autopp/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/steep-0.51.0/lib/steep/server/master.rb:426:in `block (3 levels) in start'
  /Users/autopp/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/activesupport-7.0.2.3/lib/active_support/tagged_logging.rb:99:in `block in tagged'
  /Users/autopp/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/activesupport-7.0.2.3/lib/active_support/tagged_logging.rb:37:in `tagged'
[...]

PR #574, opened 2022-06-12
Fix: Fix langserver crashes with untitled file

sevenc-nanashi:


soutaro:
Related issue: https://github.com/soutaro/steep/issues/526

Issue #549, opened 2022-05-09
Can't type check `$stdout`/`$stderr`


myronmarston:
When I reference `$stdout` or `$stderr`, `steep check` tells me it can't detect the type:

```
elasticgraph-support/lib/elastic_graph/support/logger.rb:33:29: [error] Cannot detect the type of the expression
│ Diagnostic ID: Ruby::FallbackAny
│
└           when "stdout" then $stdout
                               ~~~~~~~

elasticgraph-support/lib/elastic_graph/support/logger.rb:34:29: [error] Cannot detect the type of the expression
│ Diagnostic ID: Ruby::FallbackAny
│
└           when "stderr" then $stderr
                               ~~~~~~~
```

Since `$stdout` and `$stderr` are built-in global variables it feels like steep should understand them automatically.  Furthermore, defining the types for these manually (e.g. via `$stdout: IO` in an rbs file) doesn't work; that changes the steep error to:

```
elasticgraph-support/sig/elastic_graph/support/logger.rbs:33:0: [error] Declaration of `$stderr` is duplicated
│ Diagnostic ID: RBS::DuplicatedDeclaration
│
└ $stderr: IO
  ~~~~~~~~~~~
```

or

```
elasticgraph-support/sig/elastic_graph/support/logger.rbs:32:0: [error] Declaration of `$stdout` is duplicated
│ Diagnostic ID: RBS::DuplicatedDeclaration
│
└ $stdout: IO
  ~~~~~~~~~~~
```

(Which of the two "is duplicated" errors I get depends on which global variable is defined first in my rbs file).

So it seems like steep already has a definition for `$stdout`/`$stderr` but doesn't use the definition when it encounters the expression.

sampersand:
While `$stdout` / `$stderr` are normally `IO` instances, you can actually set them to anything which has `.write` defined:
```ruby
class MyStdout
   def write(*) = nil # swallow
end
```
Since they're global variables, they can be assigned by any anyone—including code that steep can't monitor (eg external gems). While I'd like to see this feature, I don't really think it's possible to do if we want to typecheck valid ruby

myronmarston:
> While `$stdout` / `$stderr` are normally `IO` instances, you can actually set them to anything which has `.write` defined:

Perhaps (are there any ruby docs that state that only a `write` method is required?) but according to the RBS files that ship with ruby, the type for these is `IO`:

https://github.com/ruby/rbs/blob/v2.4.0/core/global_variables.rbs#L174-L181

[...]

PR #579, opened 2022-06-13
Set up gvar env in methods

soutaro:
Fixes https://github.com/soutaro/steep/issues/549

### DIFF
--- a/test/type_construction_test.rb
+++ b/test/type_construction_test.rb
@@ -9176,4 +9176,44 @@ RUBY
       end
     end
   end
+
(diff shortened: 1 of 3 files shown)


test/type_construction_test.rb
```diff
@@ -9176,4 +9176,44 @@ def foo(x = 123, foo: true)
       end
     end
   end
+
+  def test_anonymouse_block_forwarding
```
myronmarston, line 9180: It looks like this test doesn't relate to anonymous block forwarding. Copy/paste error on the test method name?

PR #585, opened 2022-06-18
Fix discarding test_anonymouse_block_forwarding test case

kachick:
Because overrideng test case that added in https://github.com/soutaro/steep/pull/577

https://github.com/soutaro/steep/pull/579#discussion_r895802998


soutaro (approved):
🙏

Issue #595, opened 2022-06-23
Type inference for generic-type method is wrong when nil is part of the type


rpiotaix-tc:
I have the following RBS:
```rbs
class ConfigurationReader
  def read: [T](String key, default: T, constructor: ^(String value) -> T) -> T
    | (String key, default: String, ?constructor: nil) -> String
end

class KafkaSslConfiguration
  attr_reader crt_file: Pathname | nil

  def initialize: (crt_file: Pathname | nil) -> void
end
```
My code looks like:
```ruby
configuration = KafkaSslConfiguration.new(
  crt_file: reader.read("KAFKA_SSL_CRT_FILE", default: nil, constructor: ->(value) { Pathname.new(value) })
)
```

I was expecting this to work out of the box for two reasons:
1. We could either map T to nil with the default value provided and extend this with the return value of the constructor lambda making `T == Pathname | nil`
2. Infer T from the expected type of the initializer parameter that is `Pathname | nil` and see that `default` and the block return type match this type.

But I get this error:
```
[error] Cannot allow block body have type `::Pathname` because declared as type `T(1)`
│   ::Pathname <: T(1)
│
│ Diagnostic ID: Ruby::BlockBodyTypeMismatch
│
└             crt_file: reader.read("KAFKA_SSL_CRT_FILE", default: nil, constructor: ->(value) { Pathname.new(value) }),
                                                                                               ~~~~~~~~~~~~~~~~~~~~~~~
```

Is-there a way to achieve what I'm trying to express? (You read a String from the configuration Hash -- usially ENV -- and, if you don't provide a constructor to the method, then you get String as a type, and if you provide a constructor, the default value type and the returned value of the contructor should be of the same type)

PR #609, opened 2022-07-20
Fix type inference problem on generics

soutaro:
Fixes https://github.com/soutaro/steep/issues/595

### DIFF
--- a/lib/steep/type_construction.rb
+++ b/lib/steep/type_construction.rb
@@ -2731,6 +2731,10 @@ module Steep
         )
 
         if expected_block_type = block_constr.block_context.body_type
+          type_vars = expected_block_type.free_variables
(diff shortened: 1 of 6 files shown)


lib/steep/type_construction.rb
```diff
@@ -2731,6 +2731,10 @@ def type_lambda(node, params_node:, body_node:, type_hint:)
         )
 
         if expected_block_type = block_constr.block_context.body_type
+          type_vars = expected_block_type.free_variables
```
soutaro, line 2734: This is the fix.

The type error is caused by type checking lambda return type with given hint, which was type variable.
The type variable in the hint doesn't make any sense, because the lambda will be type checked with the hint in next step. The change apply a substitution that replaces all type variables to `untyped`.

PR #637, opened 2022-08-23
Support type checking block/proc self type binding

soutaro:
See https://github.com/ruby/rbs/pull/1078.

**New subtyping rules:**

* `^(...) [self: T] -> S <: ^(...) [self: T'] -> S` ⇐ `T' <: T` (`self` is essentially a parameter.)
* `^(...) [self: top] -> S <: ^(...) -> S` (Having `top` can be used as _no self binding_.)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,6 @@ group :stackprof, optional: true do
   gem "stackprof"
 end
 gem 'minitest-slow_test'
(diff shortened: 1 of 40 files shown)


soutaro:
CI is failing because of https://github.com/ruby/rbs/issues/1096.

PR #650, opened 2022-10-04
Add `steep checkfile` command

soutaro:
Spec: https://hackmd.io/8oBsfmIjQGKRCOJnYdzbvg?both


steep.gemspec
```diff
@@ -36,4 +36,5 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "rbs", ">= 2.7.0.pre"
   spec.add_runtime_dependency "parallel", ">= 1.0.0"
   spec.add_runtime_dependency "terminal-table", ">= 2", "< 4"
+  spec.add_runtime_dependency "securerandom", ">= 0.1"
```
rpiotaix-tc, line 39: @soutaro Isn't `securerandom` part of standard Ruby ? Why is this dependency necessary?

soutaro: @rpiotaix-tc It's a default gem since Ruby 3.0, which is a *gem* and installed by default. So, you are right that we can skip writing this line.

I'm trying to help `rbs collection`, which detects dependencies automatically from bundled gems. (The switching to `rbs collection` is not finished on Steep repo.)

rpiotaix-tc: @soutaro Seems to me it is included in ruby 2.6.6 (from a fresh installation using rbenv):
image

soutaro: @rpiotaix-tc It was a part of standard libraries before 3.0, not a gem. This is unnecessary, it works completely without the line.

Adding the dependency is to be more explicit and helps `rbs collection`.

PR #662, opened 2022-10-21
Fix type narrowing on case-when

soutaro:
Add support for the case that more specific class is given to `when` clause.

```rb
# @type var object: Numeric | String
object = ""

case object
when Integer
  object + 1    # Make object be Integer here
else
  object        # object is Numeric | String because the when clause above cannot delete the Numeric case
end
```


test/type_construction_test.rb
```diff
@@ -10081,4 +10081,34 @@ def test_block_splat_alias
...
   end
+
+  def test_type_narrowing_subclass
+    with_checker(<<-RBS) do |checker|
+class SuperClass
+end
+
+class SubClass < SuperClass
+end
+      RBS
```
myronmarston, line 10092: What do `SuperClass` and `SubClass` have to do with the rest of the test? It doesn't look like they are referenced or used.

myronmarston (approved):
Thanks for fixing this promptly!

test/type_construction_test.rb
soutaro: Thanks! 🙏

PR #668, opened 2022-11-16
Fixes for some RBS errors within steep gem

amatsuda:
While running `steep check` on my app, I got some errors raised from inside Steep gem. I soon learned from the documentation that I could mute these errors by adding `ignore` entry in my rbs_collection.yaml, but anyway it'd be better for steep not to raise when being type checked, so here are my quick fixes for two of them.

I'm sorry though, I couldn't fix all the errors. I'm still seeing the following error, but I'll leave it untouched because I'm not sure what is the proper fix.

```
.../steep/sig/steep/subtyping/check.rbs:100:44: [error] Type application of `::Steep::Subtyping::Relation` doesn't satisfy the constraints: (::Steep::Interface::Block | nil) <: ::Object
│ Diagnostic ID: RBS::UnsatisfiableTypeApplication
│
└       def expand_block_given: (Symbol name, Relation[Interface::Block?] relation) -> (Relation[Interface::Block] | true | Result::t)
```

### DIFF
--- a/sig/steep/diagnostic/ruby.rbs
+++ b/sig/steep/diagnostic/ruby.rbs
@@ -384,11 +384,11 @@ module Steep
       class IncompatibleAnnotation < Base
         attr_reader result: Subtyping::Result::Base
 
-        attr_reader relation: Subtyping::Relation
+        attr_reader relation: Subtyping::Relation[untyped]
 
         attr_reader var_name: Symbol
 
-        def initialize: (node: Parser::AST::Node, var_name: Symbol, result: Subtyping::Result::Base, relation: Subtyping::Relation) -> void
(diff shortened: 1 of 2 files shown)


sig/steep/diagnostic/ruby.rbs
```diff
@@ -384,11 +384,11 @@ module Steep
       class IncompatibleAnnotation < Base
         attr_reader result: Subtyping::Result::Base
 
-        attr_reader relation: Subtyping::Relation
+        attr_reader relation: Subtyping::Relation[untyped]
```
amatsuda, line 387: Maybe this can be typed stricter? Tell me if so.

soutaro: It may be... I'm not sure what the type should be. 🥲

soutaro (approved):
Thanks!

PR #703, opened 2023-01-15
Fix errors caused by non-ascii variable names

ksss:
In Ruby, even non-ascii characters can be used as local variables.

```
変数 = 1
```

However, when this code is checked with steep, it results in an `UnexpectedError`.

```
sample.rb:1:0: [error] UnexpectedError: 変数 is not a local variable
│ Diagnostic ID: Ruby::UnexpectedError
│
└ 変数 = 1
  ~~~~~~
```

This patch fixes it so that a variable with non-ascii characters does not result in an UnexpectedError.

### DIFF
--- a/lib/steep/type_inference/type_env.rb
+++ b/lib/steep/type_inference/type_env.rb
@@ -308,7 +308,13 @@ module Steep
       end
 
       def local_variable_name?(name)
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks! 🎉 🙏

PR #727, opened 2023-03-05
fix(deps): update rbs versions

chozandrias76:


chozandrias76:
I'm not sure this is expected though:
```
steep check smoke/dstr/a.rb
# Type checking files:

...........................................................................................................................................................................................................................................................................................F........................................................................

sig/steep/subtyping/check.rbs:100:44: [error] Type application of `::Steep::Subtyping::Relation` doesn't satisfy the constraints: (::Steep::Interface::Block | nil) <: ::Object
│ Diagnostic ID: RBS::UnsatisfiableTypeApplication
│
└       def expand_block_given: (Symbol name, Relation[Interface::Block?] relation) -> (Relation[Interface::Block] | true | Result::t)
                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

chozandrias76:
I was able to resolve one test error, but I'm still getting a ton when I try:
`steep check $(ls | grep 'smoke\/.*')` I could use some advise @soutaro

soutaro:
Thank Colin,

The `Relation[Interface::Block?]` is the expected type. But, Steep currently doesn't allow it...

Closed this PR because I've updated the `update-steep` branch and merged it.

PR #732, opened 2023-03-08
Require rbs-2.8.x

soutaro:
RBS 3.0 was out, but Steep 1.3 doesn't work with the major update.


Gemfile.lock
```diff
@@ -13,7 +13,7 @@ PATH
       parser (>= 3.1)
       pathname (>= 0.2.1)
       rainbow (>= 2.2.2, < 4.0)
-      rbs (>= 2.8.0)
+      rbs (> 2.8.0, <= 2.8.1)
```
soutaro, line 16: rbs-2.8.1 is the most compatible version of 2.8.x with Steep. `>= 2.8.2` has a regression that fails CI.

soutaro:
Using rbs-2.8.1 for CI because `>= 2.8.2` has a regression that fails CI.

PR #770, opened 2023-04-30
Should escape underscore for method name

ksss:
Fixed a display error when a method name is sandwiched between `_` when hovering over a method.


soutaro (approved):
👏

PR #775, opened 2023-05-14
Update RBS to 3.0

tk0miya:
Steep-1.4.0 contains signatures using "use" directive.  For that reason, steep can't work with rbs-2.x now.  So the dependency for RBS should be updated.


soutaro (approved):
🙇‍♂️

Issue #787, opened 2023-05-23
Add *no report* template to `Diagnostic::Ruby 


soutaro:

tk0miya:
In my eyes, the strict and lenient templates are very similar, and it's hard to understand their difference. Additionally, the default settings are much stricter than the strict template. So I feel the "strict" template is not strict. +1 for updating it (or both of them).

soutaro:
Updated in https://github.com/soutaro/steep/pull/871.

PR #799, opened 2023-05-23
Suppress UnsupportedSyntax warnings by default

tk0miya:
UnsupportedSyntax warnings are meaningless message for users.  So they should be suppressed by default.

refs: #787


soutaro (approved):
👍

PR #800, opened 2023-05-23
Proposal: Add Steep::Diagnostic::Ruby.silent template to suppress all warnings

tk0miya:
To use steep for completions and navigations, this adds a new settings template `Steep::Diagnostic::Ruby.silent` that suppresses all warnings.

I'm not sure if adding a new template is the best way to implement completions-only mode. But it's a simple and quick way to do that.

refs: #787


soutaro (approved):
👏

PR #815, opened 2023-06-06
Make `nth_ref` and `:back_ref` nodes optional

soutaro:


soutaro:
But, do we really want `$1` optional???

PR #827, opened 2023-07-01
completion: Support completion for optional chaining (&.)

tk0miya:
The optional chaining allows to developers to call method only if the receiver is not nil.  Therefore, CompletionProvider should consider the receiver is not nil if the optional chaining operator is given.

This fix unwraps the Optional[T] from receiver_type on completion for the optional chaining.  It will provide better completion for developers.

### DIFF
--- a/lib/steep/services/completion_provider.rb
+++ b/lib/steep/services/completion_provider.rb
@@ -665,9 +665,11 @@ module Steep
 
       def unwrap_optional(type)
         if type.is_a?(AST::Types::Union) && type.types.include?(AST::Builtin.nil_type)
-          type.types.reject! { |t| t == AST::Builtin.nil_type }
+          types = type.types.reject { |t| t == AST::Builtin.nil_type }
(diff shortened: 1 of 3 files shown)


lib/steep/services/completion_provider.rb
```diff
@@ -364,6 +384,18 @@ def items_for_trigger(position:)
           # foo::← ba
           items.push(*items_for_colon2(position: position))
 
+        when node.type == :csend && at_end?(position, of: (_ = node.loc).dot) && (_ = node.loc).dot.source == "&."
```
soutaro, line 387: Seems like the last `source` check is unnecessary because `:csend` always has `&.` part, unlike `.` in `:send` nodes. (`1 + 2`)

soutaro (changes requested):
Thank you for opening this PR.
Posted some suggestions.

soutaro (approved):
👍

PR #829, opened 2023-07-02
signature helps are not shown if the target code has comments

tk0miya:
SignatureHelpProvider fails to detect the cursor position correctly because it uses source code from AST node instead of original source code.  Therefore it misdetects the cursor position if the source code contains comments.  As a result, it does not show signature helps it the source code contains comments.

### DIFF
--- a/lib/steep/services/signature_help_provider.rb
+++ b/lib/steep/services/signature_help_provider.rb
@@ -11,10 +11,10 @@ module Steep
         subtyping.factory.env
       end
 
-      def initialize(source:, content:, subtyping:)
+      def initialize(source:, subtyping:)
         @source = source
         @subtyping = subtyping
-        @buffer = RBS::Buffer.new(name: source.path, content: content)
+        @buffer = source.buffer
       end
 
       def run(line:, column:)
--- a/lib/steep/services/signature_help_provider.rb
+++ b/lib/steep/services/signature_help_provider.rb
@@ -11,15 +11,10 @@ module Steep
         subtyping.factory.env
       end
 
-      def initialize(source:, subtyping:)
+      def initialize(source:, content:, subtyping:)
         @source = source
         @subtyping = subtyping
-
-        text =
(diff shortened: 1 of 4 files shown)


lib/steep/services/signature_help_provider.rb
```diff
@@ -11,15 +11,10 @@ def env
         subtyping.factory.env
       end
 
-      def initialize(source:, subtyping:)
+      def initialize(source:, content:, subtyping:)
```
tk0miya, line 14: I don't know adding a new argument here is a better way. It seems also better to add a `content` attribute to `Steep::Source` object. Please let me know your opinion.

lib/steep/services/signature_help_provider.rb
```diff
@@ -11,15 +11,10 @@ def env
...
         @source = source
         @subtyping = subtyping
-
-        text =
-          if source.node
-            source.node.loc.expression.source
-          end
-        @buffer = RBS::Buffer.new(name: source.path, content: text || "")
+        @buffer = RBS::Buffer.new(name: source.path, content: content)
```
soutaro, line 17: It looks like we can simply use `Source#buffer` here.

```suggestion
        @buffer = source.buffer
```

soutaro (changes requested):
Thank you for the fix. It makes sense!

How about simply using the `source.buffer`? If it works, I also want to ask you to add a test case that has comments before the Ruby code.

lib/steep/services/signature_help_provider.rb
soutaro: `Source#buffer` was added very recently, weeks ago.

tk0miya: Great! It works fine!

tk0miya:
Thank you for your valuable comment. I updated this PR :-)

soutaro (approved):
🎉

PR #835, opened 2023-07-05
Reuse the latest result to keep SignatureHelp opened while typing

tk0miya:
>Is sending the latest signature help response better if SyntaxError is raised?

It seems this approach working fine to me :-)

preview


soutaro (approved):
👍

PR #850, opened 2023-07-08
Indicate the current or next argument on signature help

tk0miya:
preview

### DIFF
--- a/lib/steep/services/signature_help_provider.rb
+++ b/lib/steep/services/signature_help_provider.rb
@@ -3,7 +3,21 @@ module Steep
     class SignatureHelpProvider
       MethodCall = TypeInference::MethodCall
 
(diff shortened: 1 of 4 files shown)


lib/steep/services/signature_help_provider.rb
```diff
@@ -98,6 +133,66 @@ def signature_help_for(node, typing)
 
         [items, index]
       end
+
+      def active_parameter_for(method_type, argument_nodes, last_argument_nodes, node)
+        return unless method_type
+
+        positionals = method_type.type.required_positionals.size + method_type.type.optional_positionals.size + (method_type.type.rest_positionals ? 1 : 0) + method_type.type.trailing_positionals.size
```
tk0miya, line 140: Note: This patch did not take care of `trailing_positionals` because I don't know about it.

PR #874, opened 2023-07-14
Give an optional type hint to lhs of `||`

soutaro:


test/type_check_test.rb
```diff
@@ -1507,4 +1507,30 @@ def test_lambda__hint_is_untyped
...
+          class Hello
+            def foo: (Integer?) -> Integer
+          end
+        RBS
+      },
+      code: {
+        "a.rb" => <<~RUBY
+          class Hello
+            def foo(a)
+              a&.then {|x| x.infinite? } || -1
```
soutaro, line 1524: Steep reported a confusing type error: 

```
Cannot find compatible overloading of method `then` of type `::Integer`
```

It was because the hint given to `&.then` call is `Integer`, and it is forwarded to block typing where `x.infinite?` has `Integer?`.

PR #875, opened 2023-07-15
Update inline type comments

tk0miya:
This patch reduces the warnings from 255 to 233. Is this helpful?


lib/steep/drivers/checkfile.rb
```diff
@@ -168,8 +168,7 @@ def run
 
         Steep.logger.info { "Starting type checking: #{request_guid}" }
 
-        diagnostic_notifications = []
```
tk0miya, line 171: I removed this variable because it seems not in used.

soutaro (approved):

PR #877, opened 2023-07-19
Support type narrowing by `Module#<`

tk0miya:
Module#< and Module#<= means the argument is the ancestor of the receiver.  Thefore this allows to Steep to do narrowing the type of the receiver to the given argument.


soutaro (approved):
Thank you for the pull request! 🎉  And I'm sorry for the big delay... 🙇‍♂️

Issue #468, opened 2021-12-20
Proposal to add `expectations_path` to Steepfile


palkan:
Hey @soutaro,
Just found [this feature](https://github.com/soutaro/steep/pull/303)—really helpful!

What about making it possible to specify `expectations_path` in the `Steepfile` to always use it (without explicitly passing `--with-expectations`)?

And also, that would require adding `--without-expectations` to avoid loading it.

PR #892, opened 2023-08-06
langserver: Support --with-expectations-path option

tk0miya:
Add `--with-expectations=PATH` option to the langserver subcommand. It suppresses warnings and info messages using expectations file like check command does.  And it helps to users to develop RBS files gradually.

### DIFF
--- a/lib/steep/server/worker_process.rb
+++ b/lib/steep/server/worker_process.rb
@@ -18,7 +18,7 @@ module Steep
         @index = index
       end
 
-      def self.start_worker(type, name:, steepfile:, steep_command:, index: nil, delay_shutdown: false, patterns: [])
+      def self.start_worker(type, name:, steepfile:, steep_command:, index: nil, delay_shutdown: false, patterns: [], with_expectations_path: nil)
         begin
           unless steep_command
             fork_worker(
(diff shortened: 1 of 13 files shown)


tk0miya:
With this patch, many warnings/infos for the Steep project are suppressed on my VSCode.

Before:
スクリーンショット 2023-08-07 3 23 27

After:
スクリーンショット 2023-08-07 3 31 40

Some warnings are still shown. But it must be helpful to typing gradually :-)


Note: I noticed https://github.com/soutaro/steep/issues/468 had been filed after I created this patch. I'm unsure which is better to enable the "expectations" mode in the langserver. Please let me know if you feel other way is better. Thanks,

lib/steep/server/worker_process.rb
```diff
@@ -41,12 +42,13 @@ def self.start_worker(type, name:, steepfile:, steep_command:, index: nil, delay
             steep_command: steep_command || "steep",
             index: index,
             delay_shutdown: delay_shutdown,
-            patterns: patterns
+            patterns: patterns,
+            with_expectations_path: with_expectations_path
```
tk0miya, line 46: Note: I did not test the spawn_worker locally. How do I test this?

tk0miya:
Now, I'm using `steep:ignore` comment instead of the expectations file. So I'm withdrawing this.
Thanks,

PR #893, opened 2023-08-07
refactor: Rename driver objects to command

tk0miya:
I found some block arguments are wrongly named (maybe via copy-paste). It would be better to rename them all to "command".


soutaro (approved):
🙏

PR #894, opened 2023-08-07
Add types for LSP::Constant::MessageType

tk0miya:
refs: https://github.com/mtsmfm/language_server-protocol-ruby/blob/main/lib/language_server/protocol/constant/message_type.rb


soutaro (approved):
👍

PR #899, opened 2023-08-17
langserver: Add --severity-level option

tk0miya:
Currently, langserver does not support `--severity-level` option. Therefore the diagnostics are different with `steep check` command by default because the result of `steep check` is filtered by severity-level warning by default.  On the other hand, the result of `steep langserver` is not filtered at all.

This adds `--severity-level` option to the `steep langserver` command to control the severity level through the option.  And it defaults to "warning" level as same as other subcommand does.

### DIFF
--- a/lib/steep/drivers/checkfile.rb
+++ b/lib/steep/drivers/checkfile.rb
@@ -105,6 +105,7 @@ module Steep
           steepfile: project.steepfile_path,
           args: [],
           delay_shutdown: true,
+          severity_level: nil,
           steep_command: jobs_option.steep_command,
           count: count
         )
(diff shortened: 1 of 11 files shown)


lib/steep/drivers/checkfile.rb
```diff
@@ -105,6 +105,7 @@ def run
           steepfile: project.steepfile_path,
           args: [],
           delay_shutdown: true,
+          severity_level: nil,
```
tk0miya, line 108: It's also better to add the `--severity-level` option to checkfile subcommand too. I'll work it after this PR merged.

tk0miya:
My proposal is satisfied by https://github.com/soutaro/steep-vscode/pull/247. So I'm withdrawing this now. Thanks!

PR #901, opened 2023-08-22
doc: Add declaration of method annotations

tk0miya:
How about adding a declaration of method annotations? I'm not sure "method annotations" is the correct word. Please let me know your thoughts.


tk0miya:
Now I added `deprecated` annotation.

Issue #472, opened 2021-12-25
Steep doesn't handle Object#present? correctly


mtsmfm:
Full demo: https://github.com/mtsmfm/steep-with-present

```ruby
require "bundler/setup"
require "active_support/all"

class Foo
  def foo
    a = ["a", nil][0]

    if a.present?
      p a * 2
    end
  end
end

Foo.new.foo

__END__
$ ruby lib/foo.rb
"aa"
```

```
$ bundle exec steep check
# Type checking files:

................................................................................F...

lib/foo.rb:9:10: [error] Type `(::String | nil)` does not have method `*`
│ Diagnostic ID: Ruby::NoMethod
│
└       p a * 2
            ~

Detected 1 problem from 1 file
```

Is it intentional?

I think it should behave like `if a`.

ksss:
@mtsmfm Hi.

Sorry for cutting in, but my understanding is that I don't think it is easy to solve this problem.

Steep has special support for some methods, a feature called *logic*.
Logic feature now defined `nil?`, `is_a?`, `kind_of?`, `instance_of?`, `!` and `===`.
This feature makes it possible to narrow down the type below the if statement.

https://github.com/soutaro/steep/blob/89d42e06f5a2c77c55756c97a40e1a6c4b8fb3a4/lib/steep/type_inference/logic_type_interpreter.rb#L16-L30

https://github.com/soutaro/steep/blob/89d42e06f5a2c77c55756c97a40e1a6c4b8fb3a4/lib/steep/ast/types/factory.rb#L385-L435

In order to solve this problem, `Object#present?` needs to support on logic,
but I think Steep needs an extension that can add arbitrary methods to the logic.

mtsmfm:
As far as I know rbs doesn't have any specification how narrowing on control flow.

I think what we need is user defined type guard on rbs.

https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

dorner:
Bumping this. The (rather confusing and minimal) readme seems to indicate that using `@type` can be used to coerce a variable, but this doesn't seem to work:

```ruby
    current_assignment = Assignment.
      find_by(merchant_id: self.merchant_id) # this returns [ Assignment | nil ]

    return if current_assignment.nil? # steep doesn't seem to realize that this narrows down to Assignment
    # @type var current_assignment: Assignment

    errors.add(:base, "#{current_assignment.merchant.name} is already assigned!")
```

Leaving out the `@type` produces an error:

[...]

PR #905, opened 2023-09-01
Proposal: Add method annotations: primitive:nil? and primitive:not_nil?

tk0miya:
To represent methods like `Object#blank?` and `Object#present?` in ActiveSupport, this adds method annotations `primitive:nil?` and `primitive:not_nil?`.

They will be considered as same as `nil?` and not `nil?` internally.

refs: #472

Please let me know your opinion about implementation, the naming of annotations, and so on.

### DIFF
--- a/test/type_construction_test.rb
+++ b/test/type_construction_test.rb
@@ -6152,6 +6152,46 @@ e + 1
     end
   end
 
+  def test_logic_receiver_is_nil_via_annotation
+    with_checker(<<-RBS) do |checker|
+      class Object
+        %a{primitive:nil?}
+        def blank?: () -> bool
+      end
+    RBS
+      source = parse_ruby(<<-RUBY)
+a = [1].first
+return if a.blank?
+a + 1
+      RUBY
+
+      with_standard_construction(checker, source) do |construction, typing|
+        construction.synthesize(source.node)
+        assert_no_error typing
+      end
+    end
+  end
+
+  def test_logic_receiver_is_not_nil_via_annotation
+    with_checker(<<-RBS) do |checker|
+      class Object
+        %a{primitive:not_nil?}
+        def present?: () -> bool
+      end
+    RBS
+      source = parse_ruby(<<-RUBY)
+a = [1].first
+return unless a.present?
+a + 1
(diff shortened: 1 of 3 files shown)


ParadoxV5:
[Someone in Ruby Discord](https://ptb.discord.com/channels/518658712081268738/1143338778384146574/1167237039134085130) asked if we can write “truthy”, i.e. exact opposite of `false | nil`. Would that in the RBS core be more preferred?

tk0miya:
The goal of this PR is to provide type narrowing to Steep. So this is not related to the "truthy" type that you mentioned.

For example, the following code is available in Ruby on Rails:

```
name = params[:name]  #=> String?
if name.present?
  name  #=> String
end
```

But current steep does not support this type narrowing.

This PR adds a new feature to define type narrowing methods via `primitive:nil?` and `primitive:not_nil? annotation.

derikson:
If we defined `#present?` as:
```rbs
class Object
  %a{pure} def present?: () -> bool
end

class NilClass
  %a{pure} def present?: () -> ::FalseClass
end
```

In the example below, shouldn't Steep have enough information to narrow `name` from `String?` to `String` inside the conditional?

```ruby
name = params[:name]  #=> String?
if name.present?
  name  #=> String
end
```

This doesn't appear to work currently, but it seems like it should.

test/type_construction_test.rb
```diff
@@ -6152,6 +6152,46 @@ def test_logic_receiver_is_nil
     end
   end
 
+  def test_logic_receiver_is_nil_via_annotation
+    with_checker(<<-RBS) do |checker|
+      class Object
+        %a{primitive:nil?}
```
mtsmfm, line 6158: Unlike `nil?`, `blank?` doesn't guarantee the receiver is `nil` (e.g. it returns `true` for empty string)

tk0miya:
Now https://github.com/soutaro/steep/pull/1497 was merged. The large motivation for this PR has already been resolved.
So I'm withdrawing this.

Additionally, https://github.com/soutaro/steep/pull/1501 will be a successor of this.

Thanks,

Issue #933, opened 2023-10-19
Improve text synchronization


soutaro:
* [x] Run type check on `textDocument/didChange` for quicker response and manual save setup
* [x] Reset file content on `textDocument/didOpen`
* [x] Set up file watchers to get notified changes of non-opened files

PR #936, opened 2023-10-20
Set up file watcher

soutaro:
Use LSP file watcher protocol for edits made by other tools.

The pattern syntax is [defined here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#pattern). It looks like compatible with Ruby's glob syntax.

#933

### DIFF
--- a/lib/steep/server/type_check_worker.rb
+++ b/lib/steep/server/type_check_worker.rb
@@ -74,11 +74,6 @@ module Steep
         when "textDocument/didChange"
           collect_changes(request)
 
-        when "textDocument/didOpen"
-          uri = request[:params][:textDocument][:uri]
-          text = request[:params][:textDocument][:text]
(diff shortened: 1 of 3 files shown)


lib/steep/server/type_check_worker.rb
```diff
@@ -74,9 +74,9 @@ def handle_request(request)
         when "textDocument/didChange"
           collect_changes(request)
 
-        when "textDocument/didOpen"
-          uri = request[:params][:textDocument][:uri]
-          text = request[:params][:textDocument][:text]
+        when "$/file/reset"
```
soutaro, line 77: Introduce new `$/file/reset` endpoint that resets the content of a file.
Used by `textDocument/didOpen` and `workspace/didChangeWatchedFiles`.

PR #938, opened 2023-10-22
Proposal: Add a new diagnostic type: NoMethodByNil

tk0miya:
Add a new diagnostic type; NoMethodByNil, that is emitted to no method calls for the objects having optional types.  It's separated from NoMethod to control the visiblity of these diagnostics individually.

It's useful to find the methods not having appropriate types on checking the Ruby application by filtering NoMethod diagnostics.

### DIFF
--- a/lib/steep/diagnostic/ruby.rb
+++ b/lib/steep/diagnostic/ruby.rb
@@ -303,6 +303,30 @@ module Steep
         end
       end
 
+      class NoMethodByNil < Base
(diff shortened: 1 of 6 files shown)


tk0miya:
`implicitly_returns_nil!` was introduced since 1.9.1 (ref: https://github.com/soutaro/steep/pull/1396).
This diagnostic type is no longer needed. Withdrawing.

PR #962, opened 2023-11-17
Ignore smoke tests when packaging gem

p0deje:
It is not necessary for runtime and makes it impossible to use the gem in symlink environments because of a space in "hello world.rb"


soutaro (approved):
👍

PR #964, opened 2023-11-19
Support steep:ignore directive

tk0miya:
Support `steep:ignore` directive via comments in Ruby source code.
As a first step, this adds `steep:ignore` and `steep:ignore end`
directives that are able to suppress diagnostic reports for the Ruby
codes between them.
    
This implementation is based on https://hackmd.io/3-01SXMLQBm8sVHWgyq9pw.

Note: I'll work on other directive options (ex. `# steep:ignore +1:+3 ImplicitAny!`) if accepted.


lib/steep/typing.rb
```diff
@@ -41,7 +41,7 @@ def initialize(source:, root_context:, parent: nil, parent_last_update: parent&.
     end
 
     def add_error(error)
-      errors << error
+      errors << error unless source.directive_map.ignored?(error)
```
tk0miya, line 44: I need to know where is the best place to suppress the errors by directive comments.
Is there better one?

soutaro:
@tk0miya I've implement this myself at #1034, and then closing this.

PR #968, opened 2023-11-24
Default to check the entire current directory when there is no `Steepfile`

ParadoxV5:
This allows Steep to work out-of-the-box with no further setup required.
Though of course users are encouraged to set up their `Steepfile` according to their project structure.

May relate to https://github.com/soutaro/steep-vscode/pull/227#issuecomment-1630886992

#### Tests
* [x] Integration Test: no idea where to put it <.<
  * [x] DriverHelper: no unit test
  * [x] Project: empty unit test
  * [x] WorkerProcess: no unit test

### Q: Why current directory (rather than `lib` + `sig`)?
This matches the minimum (i.e., least set up) project structure of dumping everything in the project root.

## Recommended follow-up

* Utilize `nil`able `Steepfile` path in tests instead of leaving the `Steepfile` non-existent.

PR #969, opened 2023-11-25
Ignore the `rbs_collection.lock.yaml` file if it’s missing or has syntax error

ParadoxV5:
By warning the problem instead of raising exceptions, it allows Steep to not crash (read: fail safe), especially when as a passive LSP.

The check on the RBS Collection configs lockfile (default: `./rbs_collection.lock.yaml`) is now per-target in the DSL rather than batch-processed after loading it.
The DSL actually loads the YAML before the batch can verify it. If the `rbs_collection.yaml` file exists but not the lockfile, this can crash in `Errno::ENOENT: No such file or directory @ rb_sysopen - path/to/rbs_collection.lock.yaml` before the batch can provide an informative error message.

#### Tests
* [x] Integration Test: in `steepfile_test.rb`
  * [x] DSL: functionality tested via Integration; `warnings` kwarg not tested
  * [x] DriverHelper: empty unit test

### DIFF
--- a/lib/steep/drivers/worker.rb
+++ b/lib/steep/drivers/worker.rb
@@ -21,7 +21,7 @@ module Steep
 
       def run()
         Steep.logger.tagged("#{worker_type}:#{worker_name}") do
-          project = load_config()
+          project = load_config(warnings: false) # Main process has already emitted any warnings
 
           reader = LanguageServer::Protocol::Transport::Io::Reader.new(stdin)
           writer = LanguageServer::Protocol::Transport::Io::Writer.new(stdout)
--- a/test/steepfile_test.rb
+++ b/test/steepfile_test.rb
@@ -188,14 +188,20 @@ EOF
       current_dir.join('rbs_collection.yaml').write('')
       project = Project.new(steepfile_path: current_dir + "Steepfile")
 
-      assert_output nil, /rbs collection/ do
-        Project::DSL.parse(project, <<~EOF)
+      begin
+        Steep.log_output = StringIO.new
+
+        Project::DSL.parse(project, <<~RUBY)
           target :app do
           end
-        EOF
-      end
+        RUBY
 
-      assert_nil project.targets[0].options.collection_lock
+        assert_match /rbs collection/, Steep.log_output.string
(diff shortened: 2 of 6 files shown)


lib/steep/drivers/worker.rb
```diff
@@ -21,7 +21,7 @@ def initialize(stdout:, stderr:, stdin:)
 
       def run()
         Steep.logger.tagged("#{worker_type}:#{worker_name}") do
-          project = load_config()
+          project = load_config(warnings: false) # Main process has already emitted any warnings
```
ParadoxV5, line 24: See https://github.com/soutaro/steep/pull/968/files#r1403807513

The warning systems of these two PRs may conflict.

ParadoxV5: Actually, these two aren’t the only places where Steep LSP crashes, despite having opportunities to fail safely.
At least These two are (likely) the easiest to trip over.

test/steepfile_test.rb
```diff
@@ -183,6 +183,39 @@ def test_collection_explicit
...
+        Project::DSL.parse(project, <<~EOF)
+          target :app do
+          end
+        EOF
+      end
+
+      assert_nil project.targets[0].options.collection_lock
+    end
+  end
+
```
ParadoxV5, line 218: There’s a good chunk of WET in this file. Other questionable choices:
* Giving up indents rather than utilizing `<<~`
* Using `#tap` instead of `#then` despite the `void` return

My tests also check the IO output since – how else do I `assert` a warning `¯\_(ツ)_/¯`

soutaro (approved):

ParadoxV5:
I didn’t check existing tests until now because I thought my patches are independent. But nope, existing tests fail in `RBS::Collection::Config::Lockfile#check_rbs_availability!` due to moving the `check_rbs_availability!` check to a place they don’t skip over.

What to do?

soutaro:
@ParadoxV5 I tried to rebasing and adding some tests this PR, but finally concluded in having another PR for this feature. #987 So, closing this PR.

PR #975, opened 2023-12-06
Handle absolute path

lloeki:
During development of [this](https://github.com/dense-analysis/ale/pull/4671) I found out that this would work:

```
# Steepfile in /some/project/path
bundle exec steep check lib/foobar.rb
```

but this would fail:

```
# Steepfile in /some/project/path
bundle exec steep check /some/project/path/lib/foobar.rb
```

The pattern being relative, it had no chance to match the absolute path provided as argument.

Fixed by making absolute path relative to base_dir before checking against pattern.

### DIFF
--- a/lib/steep/services/file_loader.rb
+++ b/lib/steep/services/file_loader.rb
@@ -15,8 +15,9 @@ module Steep
             absolute_path = base_dir + path
 
             if absolute_path.file?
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks! 🙏

PR #976, opened 2023-12-06
Stop MethodDefinitionMissing for attr_* and aliases

tk0miya:


lib/steep/type_construction.rb
```diff
@@ -3465,6 +3466,24 @@ def type_send(node, send_node:, block_params:, block_body:, unwrap: false, tapp:
         end
       end
 
+      if module_context && receiver.nil?
+        case method_name
+        when :attr_reader
+          arguments.each do |arg|
+            module_context.defined_instance_methods << arg.children[0]
```
soutaro, line 3473: Looks like it's assuming the arguments are symbol literals? Should consider the case they are not.

lloeki:
Looks like it attempts to fix #1036, which I think I have working in #1037?

PR #987, opened 2023-12-28
Better recovery from collection problems

soutaro:
This PR is to let Steep continue running even if rbs-collection is not correctly set up, while the type checking will fail.

Related to https://github.com/soutaro/steep/pull/969

It includes a commit about `Steep.ui_logger` which improves message printing.

### DIFF
--- a/lib/steep/drivers/worker.rb
+++ b/lib/steep/drivers/worker.rb
@@ -21,7 +21,7 @@ module Steep
 
       def run()
         Steep.logger.tagged("#{worker_type}:#{worker_name}") do
-          project = load_config(warnings: false) # Main process has already emitted any warnings
+          project = load_config()
 
           reader = LanguageServer::Protocol::Transport::Io::Reader.new(stdin)
           writer = LanguageServer::Protocol::Transport::Io::Writer.new(stdout)
--- a/lib/steep/project/dsl.rb
+++ b/lib/steep/project/dsl.rb
@@ -202,16 +202,13 @@ module Steep
           when Pathname
             target.collection_config_path
           when nil
-            project.absolute_path(RBS::Collection::Config::PATH)
+            default = project.absolute_path(RBS::Collection::Config::PATH)
+            if default.file?
+              default
+            end
           when false
             nil
           end
-
-        if config_path && config_path.file?
-          lockfile_path = RBS::Collection::Config.to_lockfile_path(config_path)
-          content = YAML.load_file(lockfile_path.to_s)
-          collection_lock = RBS::Collection::Config::Lockfile.from_lockfile(lockfile_path: lockfile_path, data: content)
-        end
 
         Project::Target.new(
           name: target.name,
@@ -224,7 +221,7 @@ module Steep
               stdlib_root: target.stdlib_root,
               repo_paths: target.repo_paths
             )
-            options.collection_lock = collection_lock
+            options.collection_config_path = config_path
           end,
           code_diagnostics_config: target.code_diagnostics_config
         ).tap do |target|
--- a/lib/steep/project/dsl.rb
+++ b/lib/steep/project/dsl.rb
@@ -14,9 +14,7 @@ module Steep
         attr_reader :project
         attr_reader :collection_config_path
 
-        NONE = Object.new.freeze
-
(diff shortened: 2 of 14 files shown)


lib/steep/project/dsl.rb
```diff
@@ -135,8 +135,8 @@ def vendor(dir = "vendor/sigs", stdlib: nil, gems: nil)
         end
 
         def stdlib_path(core_root:, stdlib_root:)
-          @core_root = core_root ? Pathname(core_root) : core_root
-          @stdlib_root = stdlib_root ? Pathname(stdlib_root) : stdlib_root
+          @core_root = Pathname(core_root)
```
soutaro, line 129: There were type errors here.

lib/steep/project/dsl.rb
```diff
@@ -202,17 +202,14 @@ def target(name, template: nil, &block)
           when Pathname
             target.collection_config_path
           when nil
-            project.absolute_path(RBS::Collection::Config::PATH)
+            default = project.absolute_path(RBS::Collection::Config::PATH)
+            if default.file?
```
soutaro, line 206: Skip passing the default config file to `Options` if the file doesn't exist.

lib/steep/drivers/worker.rb
```diff
@@ -21,7 +21,7 @@ def initialize(stdout:, stderr:, stdin:)
 
       def run()
         Steep.logger.tagged("#{worker_type}:#{worker_name}") do
-          project = load_config(warnings: false) # Main process has already emitted any warnings
+          project = load_config()
```
ParadoxV5, line 24:
```suggestion
          project = load_config
```

lib/steep/project/dsl.rb
```diff
@@ -14,9 +14,7 @@ class TargetDSL
         attr_reader :project
         attr_reader :collection_config_path
 
-        NONE = Object.new.freeze
```
ParadoxV5, line 17: 💯

sig/steep.rbs
```diff
@@ -13,6 +20,8 @@ module Steep
 
   self.@logger: (Logger & ActiveSupport::TaggedLogging)?
 
+  self.@ui_logger: (Logger & ActiveSupport::TaggedLogging)?
+
```
ParadoxV5, line 24: 👀 never knew this is valid RBS syntax

sig/steep/project/dsl.rbs
```diff
@@ -24,9 +24,13 @@ module Steep
...
-        attr_reader collection_config_path: Pathname?
-
-        NONE: untyped
+        # Attribute to keep track of collection configuration
+        #
+        # * `Pathname` means loading the configuration from the path
+        # * `nil` means no configuration is given
+        # * `false` means rbs-collection is disabled
+        #
+        attr_reader collection_config_path: Pathname | nil | false
```
ParadoxV5, line 33:
```suggestion
        # Attribute to keep track of collection configuration
        #
        # * `Pathname` means loading the configuration from the path
        # * `nil` means no configuration is given
        # * `false` means rbs-collection is disabled
        #
        attr_reader collection_config_path: Pathname? | false
```

Helps convey that `Pathname` and `nil` are similar (both about explicitly given configs) while `false` is unique.

sig/steep.rbs
```diff
@@ -1,6 +1,13 @@
 module Steep
   VERSION: String
 
+  # `ui_logger` is a logger for user interaction messages
+  #
+  # The *main* process has the logger.
+  # The *worker* processes disables the logging through this.
+  #
+  def self.ui_logger: () -> (Logger & ActiveSupport::TaggedLogging)
+
```
ParadoxV5, line 10: I’m not certain about juggling two loggers. Perhaps it is possible to refactor things to use just one?
* By setting worker processes’ log level to `fatal`?
  * A logger has several severity levels. How many does Steep utilize?
    * Ruby’s `warn` and `-W` is enough for my projects.
* With a singleton method `warn` on Steep or on the logger object?

ParadoxV5 (approved):
🙏

sig/steep.rbs
soutaro: Steep uses most of the log levels. Setting `fatal` in workers doesn't work. Adding `warn` method may work, but I prefer having two logger, for now.

So, we can refactor later if we have any problem.

PR #1000, opened 2024-01-16
Fix setup for Ruby 3.3

soutaro:


olleolleolle (approved):
Good one. ruby/setup-ruby is amazing.

(This is a nice codebase!)

Also: PR number one thousand!

PR #1016, opened 2024-01-16
Fix workflow 2

soutaro:


soutaro:
Hmm, I this repo cannot use merge queue because I'm not an organization...

PR #1031, opened 2024-01-24
Make defined? accept any type

sw-square:
Currently the "argument" given to `defined?` is typechecked—this doesn't exactly follow how Ruby handles `defined?`. (It evaluates everything up to the last method call).

A good usecase of this is to check if a method is defined (`raise "must have #foo method" unless defined? arg.foo`) as `.respond_to?` won't work on `top`s.


soutaro:
Can you add a test case on `type_check_test.rb`

PR #1035, opened 2024-01-29
Skip pending UI jobs

soutaro:
It looks like completion requests are sent multiple times from VSCode and it waits for seconds to finish all of them, while only the last completion request makes sense.

This PR skips prior UI jobs before processing the last queued one.

* Pros: UI responsiveness improvements
* Cons: UI requests loss -- dropping completion request always makes sense, but how about SignatureHelp?

### DIFF
--- a/lib/steep/server/change_buffer.rb
+++ b/lib/steep/server/change_buffer.rb
@@ -5,13 +5,13 @@ module Steep
       attr_reader :buffered_changes
 
       def push_buffer
-        @mutex.synchronize do
+        mutex.synchronize do
           yield buffered_changes
         end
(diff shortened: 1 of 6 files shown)


ParadoxV5:
> It looks like completion requests are sent multiple times from VSCode and it waits for seconds to finish all of them, while only the last completion request makes sense.

Is this a mistake on VSCode’s side? If it’s intended, why?

soutaro:
I think it's intended. Typing in the editor requires new completion items because the text is changed. The root cause of the unresponsive UI is the slow type checking performance of Steep.

PR #1055, opened 2024-02-21
Abort if subtyping fails to (possible) infinite loop

soutaro:
Type checking `"" =~ ""` doesn't stop by infinitely recursive subtyping check.

```
# Simplified RBS excerpt

interface _MatchAgainst[T]
  def =~: (String) -> T
end

class String
  def =~: (Regexp) -> Integer?
        | [T] (_MatchAgainst[T]) -> T
end

# Ruby code

"" =~ ""

# Subtyping trace

String <: _MatchAgainst[T_1]                           # T_1 is a fresh type variable
  =~
    [T] (_MatchAgainst[T]) -> T <: (String) -> T_1     # Expand =~ method types
      (_MatchAgainst[T_2]) -> T_2 <: (String) -> T_1   # Expand generics with fresh type variable T_2
        String <: _MatchAgainst[T_2]                   # The first subtyping relationship but with T_2, and continues
        T_2 <: T_1
```

Instead of causing `SystemStackError`, let subtyping check fail with new `LoopAbort` error.

### DIFF
--- a/lib/steep/subtyping/check.rb
+++ b/lib/steep/subtyping/check.rb
@@ -1,6 +1,8 @@
 module Steep
   module Subtyping
     class Check
+      ABORT_LIMIT = ENV.fetch("STEEP_SUBTYPING_ABORT_LIMIT", 50).to_i
+
       attr_reader :builder
       attr_reader :cache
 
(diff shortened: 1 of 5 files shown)


lib/steep/subtyping/check.rb
```diff
@@ -1,6 +1,8 @@
 module Steep
   module Subtyping
     class Check
+      ABORT_LIMIT = ENV.fetch("STEEP_SUBTYPING_ABORT_LIMIT", 100).to_i
```
soutaro, line 4: The default limit is `100`, but you can override with `$STEEP_SUBTYPING_ABORT_LIMIT` environment variable.
Note that this is only for troubleshooting and not expecting to be configured by the user for daily uses.

soutaro:
Changed the default limit to `50` because `100` was too big for Windows.

Issue #1072, opened 2024-03-08
Minitest not being seen


wieczorek1990:
Hello,
I am getting errors type checking `wieczorek1990/pass_keyword` (https://github.com/wieczorek1990/pass_keyword).
Basically I added `steep` and wanted to use it with `minitest`.
Please help.

```
pass_keyword % ./bin/typing 
# Type checking files:

....................................................................................F.

test/test_pass_keyword.rb:5:6: [warning] Cannot find the declaration of class: `TestPassKeyword`
│ Diagnostic ID: Ruby::UnknownConstant
│
└ class TestPassKeyword < Minitest::Test
        ~~~~~~~~~~~~~~~

test/test_pass_keyword.rb:5:24: [warning] Cannot find the declaration of constant: `Minitest`
│ Diagnostic ID: Ruby::UnknownConstant
│
└ class TestPassKeyword < Minitest::Test
                          ~~~~~~~~

test/test_pass_keyword.rb:7:4: [error] Type `::Object` does not have method `refute_nil`
│ Diagnostic ID: Ruby::NoMethod
│
└     refute_nil ::PassKeyword::VERSION
      ~~~~~~~~~~

test/test_pass_keyword.rb:11:4: [error] Type `::Object` does not have method `assert_nil`
│ Diagnostic ID: Ruby::NoMethod
│
└     assert_nil Kernel.pass
      ~~~~~~~~~~

Detected 4 problems from 1 file
```

ksss:
I have looked at the repositories, you are not able to load rbs of minitest.
I would recommend using the `rbs collection`.

https://github.com/ruby/rbs/blob/master/docs/collection.md

wieczorek1990:
@ksss 

I've added `rbs collection` and now it does not see the assertion methods:

```
~/S/pass_keyword (main) [1]> bundle exec steep check
# Type checking files:

..............................................................................F..........................................................................................................................................................................................................................F...............F.................F............................................F.........

.gem_rbs_collection/concurrent-ruby/1.1/utility/processor_counter.rbs:2:2: [error] Non-overloading method definition of `physical_processor_count` in `::Concurrent` cannot be duplicated
│ Diagnostic ID: RBS::DuplicatedMethodDefinition
│
└   def self.physical_processor_count: () -> Integer
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

[...]

PR #1074, opened 2024-03-10
Should not include sig/shims on spec.files

ksss:
ref: https://github.com/soutaro/steep/issues/1072

Currently, the **parser** and **concurrent-ruby** signatures in `sig/shims` are conflicting with signatures on gem_rbs_collection.
As a stopgap measure, remove `sig/shims` from the distribution file and make it internal.
As a root solution, signatures in `sig/shims` should be merged into gem_rbs_collection.


ksss:
Oh, sig/steep dependent on sig/shims...

PR #1085, opened 2024-03-24
Fix  method call type narrowing with`||` operator

### LINKED ISSUE #1079: Ruby::ArgumentTypeMismatch Cannot pass a value of type `(::Type | nil)` as an argument of type `::Type`
## Problem
Steep ignores 3 `nil` checks on the same expression.

Offending code:

```ruby
class Event
  attr_reader :pubkey
  attr_accessor :id, :sig

  def initialize(pubkey:, id: nil, sig: nil) # <--- Note that id and sig can be nil
    @pubkey = pubkey
    @id = id
    @sig = sig
  end

  def verify_signature
    return false if id.nil? || pubkey.nil? || sig.nil? # <--- Ignored by Steep

    valid_sig?(id, pubkey, sig) # <--- pubkey and sig should not be nil here. The line above ensures that
  end

  def valid_sig?(message, public_key, signature)
    true
  end
end
```

RBS:

```rbs
class Event
  attr_reader pubkey: String
  attr_accessor id: String?  # <--- id can be nil
  attr_accessor sig: String?  # <--- sig can be nil

  def initialize: (pubkey: String, ?id: String?, ?sig: String?) -> void
  def verify_signature: -> bool
  def valid_sig?: (String, String, String) -> bool # <--- none can be nil
end
```

## Expected behaviour

No errors should be reported by Steep. The method `valid_sig?` requires non-nullable arguments. And the method `verify_signature` **ensures** that the arguments `id`, `pubkey` and `sig` are not null before calling the method `valid_sig?`.

```
$ bundle exec steep check
# Type checking files:

.....................................................................................

No type error detected. 🫖
```

## Actual behaviour

The following error is reported by Steep:

```
$ bundle exec steep check
nostr.rb:24:27: [error] Cannot pass a value of type `(::String | nil)` as an argument of type `::String`
│   (::String | nil) <: ::String
│     nil <: ::String
│
│ Diagnostic ID: Ruby::ArgumentTypeMismatch
│
└     crypto.valid_sig?(id, pubkey, sig)
                                    ~~~

```

But only when the 3 nil checks on the same expression. Fewer nil checks do not trigger the error.

## Workaround

Use 2 or fewer nil checks on the same expression.

```ruby
def verify_signature
  return false if id.nil? || pubkey.nil?
  return false if sig.nil?

  valid_sig?(id, pubkey, sig)
end
```

## Steps to reproduce

1. Clone the repo `git clone git@github.com:wilsonsilva/steep-nil-argument-type-mismatch-bug.git`
2. Run `bundle install`
3. Run `bundle exec steep check`

## Environment

- Ruby version: `3.3.0`
- Steep version: `1.6.0`
- RBS version: `3.4.4`
- OS: `macOS Sonoma 14.4 (23E214)`

A tiny sample project to reproduce the problem https://github.com/wilsonsilva/steep-nil-argument-type-mismatch-bug

soutaro:
Closes #1079 

Confirmed there is no problem with `&&` type narrowing.

### DIFF
--- a/lib/steep/type_construction.rb
+++ b/lib/steep/type_construction.rb
@@ -1813,7 +1813,7 @@ module Steep
                 .for_branch(right_node)
                 .synthesize(right_node, hint: left_truthy.type, condition: true).to_ary
 
-            right_truthy, right_falsy = interpreter.eval(env: left_falsy.env, node: right_node)
+            right_truthy, right_falsy = interpreter.eval(env: right_context.type_env, node: right_node)
 
             case
             when left_falsy.unreachable
--- a/test/type_check_test.rb
+++ b/test/type_check_test.rb
@@ -1881,4 +1881,68 @@ class TypeCheckTest < Minitest::Test
       YAML
     )
   end
+
+  def test_nilp_flow_sensitive_typing_error__or
+    run_type_check_test(
+      signatures: {
+        "a.rbs" => <<~RBS
+          class Event
+            attr_reader pubkey: String?
+            attr_reader id: String?
+            attr_reader sig: String?
+
+            def foo: (String, String, String) -> bool
+          end
+        RBS
+      },
+      code: {
+        "a.rb" => <<~RUBY
+          event = Event.new()
+
+          return if event.pubkey.nil? || event.id.nil? || event.sig.nil?
+
+          event.pubkey.strip
+          event.id.size
+          event.sig.encoding
+        RUBY
+      },
+      expectations: <<~YAML
+        ---
+        - file: a.rb
+          diagnostics: []
+      YAML
+    )
+  end
+
+  def test_nilp_flow_sensitive_typing_error__and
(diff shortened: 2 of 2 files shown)


lib/steep/type_construction.rb
```diff
@@ -1813,7 +1813,7 @@ def synthesize(node, hint: nil, condition: false)
                 .for_branch(right_node)
                 .synthesize(right_node, hint: left_truthy.type, condition: true).to_ary
 
-            right_truthy, right_falsy = interpreter.eval(env: left_falsy.env, node: right_node)
+            right_truthy, right_falsy = interpreter.eval(env: right_context.type_env, node: right_node)
```
soutaro, line 1816: We need `right_context.type_env` instead of `left_falsy.env`.
The `right_context` has purities of method calls in addition to variable typings in `left_falsy.env`.

Issue #1109, opened 2024-04-30
Wildcards in `check` directive do not expand in Steepfile


euglena1215:
Using wildcards in the `check` directive of the Steepfile does not expand the files as expected.

### Steps to Reproduce
1. Create a `Steepfile` with the following content:
    ```ruby
    target :app do
      signature 'sig'

      check 'app'
      check 'packs/*/app'
    end
    ```
2. Run the command `steep project` from the command line.

### Expected Behavior
When executing the `steep project` command, all files matching `packs/*/app` are expected to be listed under `files`.

### Actual Behavior
No files matching `packs/*/app` are displayed.

### Environment Information
- Ruby version: ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [aarch64-linux]
- Steep version: 1.6.0

PR #1113, opened 2024-05-06
Fix issue with wildcard expansion in FileLoader

euglena1215:
Fixed soutaro/steep#1109

Adjusted processing to apply `Pathname.glob` for wildcard expansion before subsequent processing. 
As a result, eliminated cases where `absolute_path` is neither a file nor a directory, and removed the unnecessary branching logic.


soutaro (approved):
🙏

Issue #1120, opened 2024-05-11
Wildcard not expanding in Steepfile, affecting `steep stats`


euglena1215:
## Summary
When running the `steep stats` command, if the `check` path specified in the `Steepfile` uses a wildcard (`*`), the wildcard is not expanded properly, and as a result, the file names do not match, which causes the `steep stats` results to not display the files.

## Steps to Reproduce
Apply the following commit to the `sample` project, and then execute `bundle exec steep stats` in the `sample` directory.

```diff
 sample/Steepfile           | 2 +-
 sample/lib/foo/bar/baz.rb  | 5 +++++
 sample/sig/foo/bar/baz.rbs | 3 +++
 3 files changed, 9 insertions(+), 1 deletion(-)

diff --git a/sample/Steepfile b/sample/Steepfile
index aeb2e4c8..0e1798fe 100644
--- a/sample/Steepfile
+++ b/sample/Steepfile
@@ -3,7 +3,7 @@ D = Steep::Diagnostic
 target :lib do
   signature "sig"

-  check "lib"                       # Directory name
+  check "lib/*/bar"                       # Directory name

   library "rbs"

diff --git a/sample/lib/foo/bar/baz.rb b/sample/lib/foo/bar/baz.rb
new file mode 100644
index 00000000..cf47d591
--- /dev/null
+++ b/sample/lib/foo/bar/baz.rb
@@ -0,0 +1,5 @@
+class Baz
+  def hello
+    'hello'
+  end
+end

diff --git a/sample/sig/foo/bar/baz.rbs b/sample/sig/foo/bar/baz.rbs
new file mode 100644
index 00000000..e4b248bf
--- /dev/null
+++ b/sample/sig/foo/bar/baz.rbs
@@ -0,0 +1,3 @@
+class Baz
+  def hello: -> String
+end
```

## Expected Result
Only the statistics for `lib/foo/bar/baz.rb` should be displayed as follows:

```
[teppei@TeppeinoMacBook-Air] ~/workspace/src/github.com/euglena1215/steep/sample
% bundle exec steep stats
# Calculating stats:

 Target  File                  Status   Typed calls  Untyped calls  All calls  Typed %
---------------------------------------------------------------------------------------
 lib     lib/foo/bar/baz.rb    success            0              0          0     100%
```

## Actual Result
However, due to the wildcard not being expanded, the file `lib/foo/bar/baz.rb` does not appear in the `steep stats` results.

```
[...]

PR #1121, opened 2024-05-11
Improve wildcard handling in pattern matching for directory paths

euglena1215:
Fixed soutaro/steep#1120

Prior to this commit, the `Pattern` class used `String#start_with?` for pattern matching when directory names were specified in the patterns array. 
This method did not support the expansion of wildcards, which led to files that should have been matched by wildcard patterns being omitted from the `steep stats` output. 
By replacing `String#start_with?` with `File.fnmatch`, this commit ensures that wildcards are correctly expanded even when matching directory patterns, thereby resolving the mismatch issue.

The tests added to `pattern_test.rb` confirm that the new pattern matching logic correctly identifies files within directories that match wildcard patterns, ensuring robustness of this fix and preventing future regressions.

### DIFF
--- a/lib/steep/project/pattern.rb
+++ b/lib/steep/project/pattern.rb
@@ -46,10 +46,9 @@ module Steep
 
       def test_string(path, patterns, prefixes)
         string = path.to_s
-        extension = path.extname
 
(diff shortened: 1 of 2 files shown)


soutaro:
Thanks! 🙏

PR #1144, opened 2024-06-02
Stop to distribute sig dir

ksss:
Retry: https://github.com/soutaro/steep/pull/1074
Fix: https://github.com/soutaro/steep/issues/1072

Some signatures in Steep's sig directory conflict with those in gem_rbs_collection, making them difficult to use.
Additionally, code that uses Steep's class/module/interface does not seem to have many use cases.
To resolve this issue, I propose excluding the `sig` directory from the steep gem package.


soutaro (approved):
🙏

PR #1147, opened 2024-06-04
Fix variable untyped hover issue in rescue clause

euglena1215:
When writing `rescue StandardError => e`, `e` should be an instance of the `StandardError` class. 
However, variables declared in rescue clauses would become untyped on hover in some cases.

```ruby
class Hello
  def do_something(x)
    x
  rescue StandardError => e
    e.message # `e` becomes untyped on hover
    e # `e` is StandardError on hover
  end
end
```

This issue is fixed by explicitly updating the context.

### DIFF
--- a/lib/steep/type_construction.rb
+++ b/lib/steep/type_construction.rb
@@ -2106,6 +2106,7 @@ module Steep
               end
 
               if body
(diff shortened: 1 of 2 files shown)


soutaro (approved):
💯

PR #1176, opened 2024-06-21
Ignore `#:` annotation on `attr_*` calls

soutaro:
This improves the compatibility with `rbs-inline`, where `attr_*` calls may be annotated for the type of the attributes.

Steep now ignores `#:` annotations given to `attr_reader`, `attr_accessor`, and `attr_writer` calls without receivers. If you really need some type annotation to your custom `attr_reader` (likes) calls, you need to add `self.` receiver or surround them with parens.

```rb
class Foo
  # Type declaration for rbs-inline, Steep ignores the annotation
  attr_reader :foo #: String

  # Steep detects the annotation because it's with `self` receiver
  self.attr_reader :foo #: String

  # Steep detects the annotation because of the parens
  (attr_reader :foo) #: String
end
```

### DIFF
--- a/lib/steep/source.rb
+++ b/lib/steep/source.rb
@@ -475,6 +475,15 @@ module Steep
               return assertion_node(node, last_comment)
             end
           else
+            if (receiver, name, * = deconstruct_send_node(node))
+              if receiver.nil?
+                if name == :attr_reader || name == :attr_writer || name == :attr_accessor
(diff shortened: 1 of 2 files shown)


soutaro:
`rbs-inline` currently uses `#::` syntax, but I will make it to `#:`. And this PR is needed to avoid the conflict.

PR #1178, opened 2024-06-23
Support literals for Rational and Complex

ksss:
Currently, steep does not support the syntax for `1r`, `1i`  and `1ri`, and they default to `untyped` type.
In this PR, I have made it possible to interpret `1r` as `Rational`, `1i` as `Complex` and `1ri` as `Complex`.


soutaro (approved):
💯

Issue #661, opened 2022-10-17
Large number of files slows down the checking process.


ksss:
I have read and attempted to steep check a large number of RBS files, over 10000.
I found that there is a big difference in execution time compared to when all the contents are in one file.
I think this is an incentive to put the RBS files in one file instead of separating them, so the difference in execution time between the two should be smaller.

gen.rb
```rb
#! /usr/bin/env ruby

require 'pathname'

split = Pathname.new('split')
split.rmtree rescue nil
split.mkdir
n = ARGV[0].to_i
(n - 1).times.map { |i| "C#{i}" }.each do |c|
  split.join("#{c}.rbs").write(<<~RBS)
    class #{c}s
    end
  RBS
end
```

How to generate RBS files.

```
$ ruby gen.rb 2000
```

How to concat RBS files.

```
$ cat split/* > concat/concat.rbs
```

How to measure time.

```
$ time bundle exec steep check
```

Measurements were taken by switching directories.

```
$ cat Steepfile
target :sample do
  check "sample.rb"
  signature "split" # or "concat"
end
```

8 cpu execution time.

file count  | split user | split real | concat user | concat real
-- | -- | -- | -- | --
2000 | 38.92 | 10.774 | 17.67 | 5.04
4000 | 75.43 | 22.585 | 22.19 | 6.612
6000 | 130.57 | 43.663 | 25.55 | 8.019
8000 | 204.71 | 72.76 | 31.27 | 9.888
10000 | 299.07 | 114.17 | 35.15 | 11.57

<img width="605" alt="image" src="https://user-images.githubusercontent.com/935310/196191297-eabdb531-8ac8-4eb2-84c2-d249c0650c4c.png">

soutaro:
Nice found...

The reason would be because Steep distributes the RBS validation tasks to worker processes by file. If we have 10,000 files and 10 workers, each worker has 100 validation tasks to execute. If we have 10 files and 10 workers, each worker has 1 validation task to execute. I guess this causes the difference, but not confirmed yet.

ksss:
Looking at the `$ htop`, it appeared that there were times when tasks were concentrated on only one cpu.
I have not been able to find out which task it is...

PR #1184, opened 2024-07-04
Improve performance when there are many files

pocke:
Probably this PR fixes https://github.com/soutaro/steep/issues/661.

The execution time of `steep check` increases with `O(n^2)` order (n == number of files).
This PR improves the performance. The order will probably be `O(n)`.

## Problem

`TypeCheckRequest#all_paths` is called on each file. This method allocates a large `Set` if `steep check` checks many files. It makes `steep check` slow.

## Solution

Avoid allocating the large set in `all_paths` method.

## Benchmarking

I benchmarked `steep check` command with the following script, creating many `.rb` and `.rbs` files.

```ruby
# Usage: ruby setup.rb 100

require 'pathname'

lib = Pathname('lib').tap { _1.rmtree; _1.mkpath }
sig = Pathname('sig').tap { _1.rmtree; _1.mkpath }

ARGV.first.to_i.times do |i|
  lib.join("foo#{i}.rb").write(<<~RUBY)
    class Foo#{i}
    end
  RUBY
  sig.join("foo#{i}.rbs").write(<<~RUBY)
    class Foo#{i}
    end
  RUBY
end
```

The result is the following.

Screen Shot 2024-07-04 at 16 32 08

https://docs.google.com/spreadsheets/d/1cF6KGS12i2_dBOW0GQ59EDEkhTnpxR-qAokEoB73VKQ/edit?usp=sharing

With this patch, the execution time looks linear. (The n=10000 case looks strange... I guess the execution time is unexpectedly long for some unintentional reason. But I'm not sure)

## Profiling

I found this problem while investigating Steep's memory usage issue. I found a bottleneck for the `Set` allocation with `memory_profiler` gem on a Rails application.

I've confirmed that this change dramatically decreases the memory allocation size for `Set` with the memory profiler.
The results are n=4000 cases.

### before

```
Total allocated: 7563837242 bytes (5427818 objects)
Total retained:  1062365 bytes (171 objects)

allocated memory by gem
-----------------------------------
7114488384  set
 253915960  pathname
  53144417  steep/lib
  47872344  language_server-protocol-3.17.0.3
  46181320  json-2.7.2
  29675372  rbs-3.5.1
  12868648  uri
   3050296  other
   2588320  activesupport-7.1.3.4
     36592  optparse
      3671  rainbow-3.1.1
      3600  securerandom-0.3.1
      2776  bundled_gems
      2001  concurrent-ruby-1.3.3
      1893  arm64-darwin21
      1448  logger-1.6.0
       200  digest
```

### after

```
Total allocated: 450887826 bytes (5314647 objects)
Total retained:  1062365 bytes (171 objects)

allocated memory by gem
-----------------------------------
 253915960  pathname
  53144417  steep/lib
  47872472  language_server-protocol-3.17.0.3
  46181320  json-2.7.2
  29675372  rbs-3.5.1
  12868648  uri
   3050296  other
   2588320  activesupport-7.1.3.4
   1538880  set
     36592  optparse
      3671  rainbow-3.1.1
      3560  securerandom-0.3.1
      2776  bundled_gems
      2001  concurrent-ruby-1.3.3
      1893  arm64-darwin21
      1448  logger-1.6.0
       200  digest
```

-----

BTW, I am currently working on reducing the peek memory usage of Steep. This change affects the execution time but probably doesn't affect the peek memory usage, unfortunately, because they're short-lived objects.

### DIFF
--- a/lib/steep/server/master.rb
+++ b/lib/steep/server/master.rb
@@ -39,7 +39,7 @@ module Steep
         end
 
         def percentage
-          checked_paths.size * 100 / all_paths.size
+          checked_paths.size * 100 / total
         end
 
         def all_paths
@@ -58,7 +58,7 @@ module Steep
         end
 
         def finished?
-          unchecked_paths.empty?
+          total <= checked_paths.size
         end
 
         def unchecked_paths


soutaro (approved):
🙏

PR #1187, opened 2024-07-08
Add Vim linting information

lloeki:
I've added basic linter-mode support for `steep` in ALE for Vim/Neovim in https://github.com/dense-analysis/ale/pull/4671 so let's document that and hopefully get a bit more `steep` adoption!

Hopefully someday I'll find a way to add proper LSP mode (which kind of works but had some brokenness last time I checked)


soutaro (approved):
Thanks!

PR #1209, opened 2024-08-18
LSP: Support formatting RBS file

tk0miya:
To keep RBS files clean, this adds a new feature to format RBS file to Steep LSP.


soutaro:
As mentioned at ruby-jp slack, the `RBS::Writer` removes some comments. Essentially, we need to fix `RBS::Writer`, but it's not trivial...

Do you think the behavior acceptable?

tk0miya:
I think users expect the formatters to be safe. So such destructive behavior is not accepted by users, including me.

So I'd like to withdraw this PR until the RBS formatters mature enough.

PR #1217, opened 2024-09-09
No `timeout: nil`

soutaro:
Test fails with `ThreadError` on `#push` with ruby-3.0 and ruby-3.1.

```
/home/runner/work/steep/steep/lib/steep/server/master.rb:1028:in `push': queue full (ThreadError)
```

This is because the `timeout:` keyword arg is introduced at Ruby 3.2, so the `push(object, timeout: nil)` is equivalent to `push(object, true)` which means raising error immediately when queue is full.

Deleting `timeout: nil` would solve the problem.


soutaro:
Ruby 3.1 will be EOL 2025-03-31.

PR #1223, opened 2024-09-15
Support optional keys in record types

soutaro:
The semantics of optional keys in record in Steep is *the hash may not have any value associated with the optional key*.

So, 

1. Assigning a hash without the optional key is allowed -- `{ id: 123 } #: { id: Integer, ?name: String }`
2. Having incompatible type with the optional key is prohibited (but, it causes unsoundness...)
3. The `#[]` operator of the record type returns an optional type, `#fetch` returns non-optional type (if the value type is not an optional)


ParadoxV5:
* ruby/rbs#504

PR #1238, opened 2024-09-30
Remove unused rules

euglena1215:
The following rules are defined but not used.

- `UnexpectedSplat`
- `IncompatibleMethodTypeAnnotation`
- `IncompatibleTypeCase`

The above rules have been removed because unused rules remain that can easily lead to misunderstandings.

You may need to keep it if you plan to use it in the future.


soutaro (approved):
Thanks!

Issue #149, opened 2020-06-06
Blocks, Procs and Methods


soutaro:
Moved from https://github.com/ruby/rbs/issues/298.

Thanks @timhatch for reporting an issue. I think it's a problem more related to Steep than RBS and moved to here. The question is the following code doesn't work well with Steep.

```rb
  def by_country(person)
    person.slice(:country, :lastname, :id).values
  end
  
  my_array.sort_by(&method(:by_country))
```

soutaro:
The problem is that the `method` method doesn't return a correct type with respect to each actual method.

It is because RBS doesn't allow writing that type. I don't think it's reasonable to extend RBS to allow that type, nor I don't think we can define a type system to support it. Then the problem is what is a better way for type checking this Ruby code.

One of the possible workarounds would be to allow using `&` operator with `Method` operand and skip some type checking.

```rb
my_array.sort_by(&method(:by_country))     # Shows some warning but type checks.
my_array.sort_by(&method(:inspect))        # This also type checks because it just skips checking!
```

Do you have any opinion?

timhatch:
Hi Soutaro

First, my apologies for raising this in rbs. I had both the Steep and rbs repositories open and posted the issue to rbs by mistake... 

re. _I don't think it's reasonable to extend RBS to allow that type, nor I don't think we can define a type system to support it._ I agree. 

Thanks for the work on Steep and rbs. I have wanted to test ruby type-checking for some time and using Steep/rbs on an existing code-base is proving really interesting. Is there a forum for general feedback/comments - I do not want to raise "Issues" unnecessarily.

timhatch:
To add to this, as the original code being type-checked is common across various modules/classes being checked, then I should note that returning a proc instead of using the `method` method also shows some peculiar behaviour, e.g.

```ruby
def by_rank
   ->(x) { [x[:rank], x[:country], x[:lastname]] }
 end
```

type-checks without problem when called in `my_array.sort_by(&by_rank)`

However the following alternatives both raise an error:
```ruby
  def by_rank
    lambda do |x| 
      [x[:rank], x[:country], x[:lastname]]
    end
  end

  def by_rank
    proc do |x| 
      [x[:rank], x[:country], x[:lastname]]
    end
  end
```

In both cases, Steep reports as follow:
`MethodBodyTypeMismatch: method=by_rank, expected=^(::Hash::object) -> ::Array[[::Integer, ::String, ::String]], actual=::Proc (def by_rank)
[...]

PR #1245, opened 2024-09-30
Raise UnsupportedSyntax if a proc object is given as block-pass-argument

tk0miya:
At present, users will see BlockTypeMismatch diagnostic when a proc object is passed as block-pass-argument.  We'll usually see the diagnostic with `&method` idiom.

```ruby
[1, 2, 3].map(&method(:puts))
```

But there is nothing to do from the user side because such a call is not invalid.  It seems like a false positive.

This changes the diagnostic type for the case to UnsupportedSyntax.  It notifies users that the source code is valid and that change is unnecessary.

refs: #149


lib/steep/type_construction.rb
```diff
@@ -4191,6 +4191,16 @@ def try_method_type(node, receiver_type:, method_name:, method_overload:, argume
                   )
                 else
                   # non-nil value is given
+                  if node_type.is_a?(AST::Types::Name::Instance) && node_type.name.to_s == '::Proc'
```
tk0miya, line 4194: Is it better to move this into `#synthesize`?

tk0miya:
Now, https://github.com/soutaro/steep/pull/1276 has been merged into the master. But this is still needed, I believe.
We still have the possibility of getting `BlockTypeMismatch` because #1276 only covers the case of `&method(:name)`.

For example, the following code will cause the diagnostic:

```
meth = method(:puts)
[1, 2, 3].map(&meth)

block = proc { |n| n.to_s }
[1, 2, 3].map(&block)
```

PR #1249, opened 2024-10-02
doc: Add diagnostics for Ruby page

euglena1215:
When running steep check, various errors are displayed, but there is no clear explanation of what these errors mean.

So I created a document for error reference.

Japanese version: https://tech.timee.co.jp/entry/2024/10/02/153330

### DIFF
--- a/guides/src/diagnostics/ruby.md
+++ b/guides/src/diagnostics/ruby.md
@@ -1,36 +1,48 @@
-# Diagnostics for Ruby
-Each error description will follow the format below.
+# Ruby Code Diagnostics
 
-````markdown
-## Error Name
+## Configuration Templates
+Steep provides several templates to configure diagnostics for Ruby code.
+You can use these templates or customize them to suit your needs via `#configure_code_diagnostics` method in `Steepfile`.
 
-**Description**: Brief error description
+The following templates are available:
 
-**Example:**
+<dl>
+<dt><code>Ruby.all_error</code></dt>
+<dd>This template reports everything as an error.
 
-```ruby
-Ruby code that detects the error
-```
+</dd>
+<dt><code>Ruby.default</code></dt>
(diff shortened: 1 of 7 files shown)


guides/src/diagnostics/ruby.md
```diff
@@ -0,0 +1,1438 @@
+# Diagnostics for Ruby
+Each error description will follow the format below.
+
+````
```
ParadoxV5, line 4:
`````suggestion
````markdown
`````

File: guides/src/diagnostics/ruby.md [hunk 2]
```diff
@@ -0,0 +1,1438 @@
...
+│
+└ lambda {|x| x + 1 } #: ^(Integer) -> String
+         ~~~~~~~~~~~~
+```
+
+**severity**:
+
+| all_error | default | strict | lenient | silent |
+| --- | --- | --- | --- | --- |
+| error | warning | error | information | nil |
```
ParadoxV5, line 83: Looks like `default` is between `strict` and `lenient`.
```suggestion
| all_error | strict | default | lenient | silent |
| --- | --- | --- | --- | --- |
| error | error | warning | information | nil |
```
Also, I’m not sure if it’s necessary to include the `all_error` and `silent` severities, as they (should) set everything to `error` and `nil` respectively.

ParadoxV5 (approved):
👍 I needed this since years ago. 🙏

guides/src/diagnostics/ruby.md
euglena1215: > Looks like `default` is between `strict` and `lenient`.

Thanks! I'll fix it.

> Also, I’m not sure if it’s necessary to include the `all_error` and `silent` severities, as they (should) set everything to `error` and `nil` respectively.

This was a point that bothered me too. I thought it would be valuable to have a list of error presets, so I'm listing them here, but I'd like to hear soutaro-san's thoughts.

PR #1270, opened 2024-10-08
Support trailing_positionals

### LINKED ISSUE #1172: Unexpected MethodParameterMismatch with positional argument after optional argument
Steep raises an error with the following RBS and Ruby files.

`test.rbs`

```rbs
class C
  def f: (?untyped a, untyped b) -> void
end
```

`test.rb`

```ruby
class C
  def f(a = nil, b)
  end
end
```

```console
$ bundle exec steep --version
1.7.1

$ bundle exec steep check
# Type checking files:

...................................F.................................................

test.rb:2:17: [error] The method parameter is incompatible with the declaration `(?untyped) -> void`
│ Diagnostic ID: Ruby::MethodParameterMismatch
│
└   def f(a = nil, b)
                   ~

```

I expect that it does not raise any error, but Steep displays an error.
The error message says the type of method `f` is `(?untyped) -> void`. Probably Steep ignores the positional argument after an optional argument.

tk0miya:
Closes #1172

### DIFF
--- a/test/type_check_test.rb
+++ b/test/type_check_test.rb
@@ -2054,6 +2054,32 @@ class TypeCheckTest < Minitest::Test
     )
   end
 
+  def test_method_def__trailing
+    run_type_check_test(
+      signatures: {
+        "a.rbs" => <<~RBS
+          class Hello
+            def f: (?untyped a, untyped b) -> void
+          end
+        RBS
+      },
+      code: {
+        "a.rb" => <<~RUBY
+          class Hello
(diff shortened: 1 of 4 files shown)


test/type_check_test.rb
```diff
@@ -2054,6 +2054,32 @@ def f(x, y, z)
     )
   end
 
+  def test_method_def__trailing
```
tk0miya, line 2057: Please let me know if there is more better place to add a test case.

PR #1276, opened 2024-10-09
Support &method(:name) call for block_pass

tk0miya:
Pass `&method(:name)` to the block argument on method call is usually used, and well-known idiom in Ruby.  Therefore it is valuable to support the idiom on type check.

This gives a special treatment to it to support type checking.

### DIFF
--- a/lib/steep/type_construction.rb
+++ b/lib/steep/type_construction.rb
@@ -2402,14 +2402,19 @@ module Steep
               type, constr = synthesize(value_node, hint: hint)
 
               if hint.is_a?(AST::Types::Proc) && value_node.type == :send && value_node.children[1] == :method && AST::Builtin::Method.instance_type?(type)
+                receiver_node = value_node.children[0] #: Parser::AST::Node?
+                receiver_type = receiver_node ? typing.type_of(node: receiver_node) : self_type
                 method_name = value_node.children[2].children[0] #: Symbol
-                if method = calculate_interface(AST::Types::Self.instance, private: true)&.methods&.[](method_name)
-                  # TODO: select appropriate method type from `method.method_types`.
-                  type = AST::Types::Proc.new(
-                    type: method.method_types.first&.type || raise,
-                    block: nil,
-                    self_type: nil
-                  )
+                if method = calculate_interface(receiver_type, private: true)&.methods&.[](method_name)
+                  if method_type = method.method_types.find {|method_type| method_type.accept_one_arg? }
+                    if method_type.type_params.empty?
+                      type = AST::Types::Proc.new(
+                        type: method_type.type,
+                        block: method_type.block,
+                        self_type: nil
+                      )
+                    end
+                  end
                 end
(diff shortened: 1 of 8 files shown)


lib/steep/type_construction.rb
```diff
@@ -2401,7 +2401,17 @@ def synthesize(node, hint: nil, condition: false)
             if value_node
               type, constr = synthesize(value_node, hint: hint)
 
-              if hint.is_a?(AST::Types::Proc) && value_node.type == :sym
+              if hint.is_a?(AST::Types::Proc) && value_node.type == :send && value_node.children[1] == :method && type.is_a?(AST::Types::Name::Instance) && type.name.to_s == '::Method'
+                method_name = value_node.children[2].children[0]
+                if method = calculate_interface(AST::Types::Self.instance, private: true)&.methods&.[](method_name)
+                  # TODO: select appropriate method type from `method.method_types`.
```
tk0miya, line 2407: I'm unsure how to choose the appropriate method type from `method.method_types`. So I use the first one.

test/type_construction_test.rb
```diff
@@ -6950,6 +6950,62 @@ def m: [T] () { (untyped) -> T } -> T
...
+r = x.map(&method(:my_to_s))
+      EOF
+
+      with_standard_construction(checker, source) do |construction, typing|
+        pair = construction.synthesize(source.node)
+
+        assert_typing_error typing, size: 1 do |errors|
+          assert_any!(errors) do |error|
+            assert_instance_of Diagnostic::Ruby::BlockTypeMismatch, error
+            assert /^\^\(::Integer\) -> X\(\d+\)$/ =~ error.expected.to_s
```
tk0miya, line 6999: I use regexp to check the BlockTypeMismatch#expected. Is there a better way?

soutaro: Ah, in fact, you may want to use the newer `type_check_test.rb`.

lib/steep/type_construction.rb
```diff
@@ -2401,7 +2401,18 @@ def synthesize(node, hint: nil, condition: false)
             if value_node
               type, constr = synthesize(value_node, hint: hint)
 
-              if hint.is_a?(AST::Types::Proc) && value_node.type == :sym
+              if hint.is_a?(AST::Types::Proc) && value_node.type == :send && value_node.children[1] == :method && AST::Builtin::Method.instance_type?(type)
+                method_name = value_node.children[2].children[0] #: Symbol
+                if method = calculate_interface(AST::Types::Self.instance, private: true)&.methods&.[](method_name)
+                  if overload = method.method_types.find {|method_type| method_type.type.params.nil? || method_type.type.params.optional? }
+                    type = AST::Types::Proc.new(
+                      type: method.method_types.first&.type || raise,
```
tk0miya, line 2409: Is it better to use `overload` here?
```suggestion
                      type: overload,
```

PR #1286, opened 2024-10-18
refactor: Use Hash#fetch instead of Hash#[] to resolve type errors

tk0miya:
135 -> 47

Note: `Hash#fetch` raises KeyError. So we need to merge this carefully. I'm okay to divide this PR to smaller ones.

### DIFF
--- a/lib/steep/services/hover_provider/ruby.rb
+++ b/lib/steep/services/hover_provider/ruby.rb
@@ -65,14 +65,14 @@ module Steep
         def method_definition_for(factory, type_name, singleton_method: nil, instance_method: nil)
           case
           when instance_method
-            factory.definition_builder.build_instance(type_name).methods[instance_method]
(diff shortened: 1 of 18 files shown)


lib/steep/services/hover_provider/ruby.rb
```diff
@@ -65,22 +65,22 @@ def project
...
           case
           when instance_method
-            factory.definition_builder.build_instance(type_name).methods[instance_method]
+            factory.definition_builder.build_instance(type_name).methods.fetch(instance_method)
           when singleton_method
             methods = factory.definition_builder.build_singleton(type_name).methods
 
             if singleton_method == :new
-              methods[:new] || methods[:initialize]
+              methods[:new] || methods.fetch(:initialize)
```
tk0miya, line 73: This is also a bit strange. The left hand is `#[]`, but the right hand is `#fetch`. It's a little difficult to understand the behavior for me.

Issue #1344, opened 2024-11-21
Run `rbs collection install` when starting type checking


soutaro:
Related to https://github.com/soutaro/steep-vscode/issues/379. (CC: @tk0miya)

PR #1345, opened 2024-11-22
Install rbs collection automatically

soutaro:
#1344 

It runs `rbs collection --frozen` (equivalent) if

* Collection lock file is generated, and
* RBS files are missing


soutaro:
Do we need an option to disable the automatic install? I had thought we need, but not sure.

If the installation process doesn't work, we can add a step to `rbs collection install` before `steep` invocation.

PR #1347, opened 2024-11-22
Expand `array(splat(expr` node

soutaro:
This happens with `*node` node, at masgn right hand side.

```rb
a, b = *regexp.match("foo")     # The expected type of a and b is `String?`, but was `String | [] | nil`
```


soutaro:
@rhiroe reported this problem. Thanks!

PR #1356, opened 2024-11-25
emit UnreachableBranch to the "void" condition

tk0miya:
It would be better to emit UnreachableBranch diagnostics to use "void" variables for if-conditions.


test/type_construction_test.rb
```diff
@@ -2983,6 +2983,75 @@ def test_if_annotation_success
     end
   end
 
+  def test_unreachable_if_void
```
soutaro, line 2986: Please move these tests to `type_check_test.rb` if you can.

tk0miya: Updated.

soutaro (approved):
🙏

PR #1360, opened 2024-11-26
Update example to not use `^` as a hash function

amomchilov:
XOR is not a good hash function, and this example might give people the wrong idea.

Better to just outsource to the built-in Hash function used by `Array`. It's more idiomatic, and will lead to less collisions.


soutaro (approved):
Thanks!

Issue #1273, opened 2024-10-09
Unexpected error with anonymous arguments forwarding


pvcresin:
I found unexpected error pattern.

### Repro

- Ruby version: ruby 3.3.3 (2024-06-12 revision f1c7b6f435) [arm64-darwin23]
- Steep version: 1.8.1

```rbs
# sample.rbs
class Sample
end
```

```rb
# sample.rb
class Sample
  def args_only(*)
    p(*)
  end

  def kwargs_only(**)
    p(**)
  end

  def block_only(&)
    ["block"].each(&)
  end

  def foo(*, **, &)
    args_only(*)
    kwargs_only(**)
    block_only(&)
  end
end

Sample.new.foo(1, a: 1) { |t| p t }
```

```console
$ ruby lib/sample.rb
1
{:a=>1}
"block"
```

<details>
<summary>$ bundle exec steep check</summary>
<pre>
$ bundle exec steep check
# Type checking files:

.............................................2024-10-09 15:14:38.113: FATAL: [Steep 1.8.1] [typecheck:typecheck@6] [background] [#typecheck_source(path=lib/sample.rb)] [#type_check_file(lib/sample.rb@lib)] [synthesize:(1:1)] [synthesize:(1:1)] [synthesize:(2:3)] [synthesize:(6:3)] [synthesize:(7:5)] [[T < ::_Inspect] (T) -> T] [synthesize:(7:7)] Unexpected error: RuntimeError
..2024-10-09 15:14:38.128: ERROR: [Steep 1.8.1] [typecheck:typecheck@6] [background] [#typecheck_source(path=lib/sample.rb)] [#type_check_file(lib/sample.rb@lib)] [synthesize:(1:1)] [synthesize:(1:1)] [synthesize:(2:3)] [synthesize:(6:3)] [synthesize:(7:5)] [[T < ::_Inspect] (T) -> T] [synthesize:(7:7)]   /Users/toriyama/.rbenv/versions/3.3.3/lib/ruby/gems/3.3.0/gems/steep-1.8.1/lib/steep/type_construction.rb:5033:in `block in type_hash'
[...]

PR #1369, opened 2024-11-29
Emit UnsupportedSyntax to forwarded_kwrestarg

tk0miya:
Close: #1273


soutaro:
The keyword arguments are handled in `SendArgs` and adding something in `TypeConstruction` would not fix the issue. I tried to add errors from `SendArgs` but couldn't find a good extension point to emit an error...

PR #1406, opened 2024-12-13
Fix steep:ignore does not work with CR/LF

### LINKED ISSUE #1376: `steep:ignore` family sometimes doesn’t work in CRLF files
```console
$ echo -e '\r\nfoo # steep:ignore' > ruby.rb
$ hexdump *
0000000 0a0d 6f66 206f 2023 7473 6565 3a70 6769
0000010 6f6e 6572 000a
0000015
$ steep check
2024-12-02 16:35:47.772: ERROR: [Steep 1.8.3] Cannot find a configuration at Steepfile: `steep init` to scaffold. Using current directory...
# Type checking files:

...........................................................F........................

ruby.rb:2:0: [error] Type `::Object` does not have method `foo`
│ Diagnostic ID: Ruby::NoMethod
│
└ foo # steep:ignore
  ~~~

```

tk0miya:
parser gem removes CR characters before parsing.  Therefore `Parser:: Source::Buffer` does not contain any CR characters.  As a result, `Parser::Source::Comment` points the location without CR characters.

On the other hand, `RBS::Buffer` contains CR characters.  This difference causes `steep:ignore` comment disabled.

This converts the location of comments using line and column

Close #1376


soutaro (approved):
👍

PR #1440, opened 2024-12-24
Fix a runtime warning for key :cursor is duplicated and overwritten on line 99

ydah:
```
❯ ruby -v
ruby 3.4.0rc1 (2024-12-12 master 29caae9991) +PRISM [arm64-darwin23]
❯ RUBYOPT=-w bundle exec rake
/ydah/steep/test/logic_type_interpreter_test.rb:99: warning: key :cursor is duplicated and overwritten on line 99
```


soutaro (approved):
🙏

Issue #1409, opened 2024-12-15
Proposal: ERB Support


tk0miya:
ERB is a widely used template library in Ruby. It is adopted as the default view library in Ruby on Rails. It is one of the popular template libraries in Ruby. I believe it's valuable to support it in Steep.

Three enhancements are necessary to let Steep support ERB. 

* Allow to change the target file extension
* Convert ERB code into Ruby before type-checking
* Allow to change the "self" context via global-level annotation

## Allow to change the target file extension

At present, Steep expects the target source code to be ".rb". It is hard-coded in some modules.

https://github.com/soutaro/steep/blob/81da4a1172fa61551ec4f0d871fb5ed7cd588d2f/lib/steep/project/dsl.rb#L105-L107
https://github.com/soutaro/steep/blob/81da4a1172fa61551ec4f0d871fb5ed7cd588d2f/lib/steep/server/master.rb#L422-L425

It would be better to specify file extension in Steepfile:

```
target :views do
  check "app/views"
  ext ".erb"
end
```

## Convert ERB code into Ruby before type-checking

ERB to Ruby code conversion is needed to type-check ERB.

Input:

```
<html>
  <body>
    <ul>
    <% languages.each do |langage| %>
      <li><%= language %>
    <% end %>
    </ul>
  </body>
</html>
```

Output: (replace HTML tags with whitespace)

```
      
        
        
       languages.each do |langage|   
              language   
       end   
         
         
       
```

## Allow to change the "self" context via global-level annotation

In the Rails application, some instance variables and helper methods are given from other components (controllers, helpers, ActionView, and so on).

Therefore, it is needed to switch the "self" context at the global level. 

```
<%# @type self: MyERBTemplate %>
<%= @some_instance_variables %>
<% some_helper_methods %>
```

It must be helpful if we'll support other DSLs.

For example, we'll also be able to support Gemfile and Raketask:

```
# @type self: Gemfile

gem 'some_gem'
gem 'another_gem'
```

```
# @type self: Raketask

desc "my task"
task :my_task do
  ...
end

namespace :new_group do
  ...
end
```

PR #1454, opened 2025-01-03
Steepfile: Allow to change the extension of the source files

tk0miya:
To support ERB, this allows to change the extension of the source files via `ext` setting in Steepfile.

Example:

```
target :views do
  check "app/views"
  ext ".erb"
end
```

refs: #1409

### DIFF
--- a/lib/steep/project/dsl.rb
+++ b/lib/steep/project/dsl.rb
@@ -74,6 +74,10 @@ module Steep
           sources.concat(args)
         end
 
+        def ext(ext)
+          @ext = ext
+        end
+
         def ignore(*args)
           ignored_sources.concat(args)
         end
@@ -103,7 +107,7 @@ module Steep
         end
 
         def source_pattern
-          Pattern.new(patterns: sources, ignores: ignored_sources, ext: ".rb")
+          Pattern.new(patterns: sources, ignores: ignored_sources, ext: @ext || ".rb")
         end
 
         def signature_pattern
--- a/lib/steep/server/master.rb
+++ b/lib/steep/server/master.rb
@@ -421,7 +421,7 @@ module Steep
                   end
                   target.source_pattern.prefixes.each do |pat|
                     path = project.base_dir + pat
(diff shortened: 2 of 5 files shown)


ParadoxV5:
Does `check 'app/views/**/*.erb'` work already?

tk0miya:
Not yet. But I'll post some PRs to support it. This is a first step for it.

lib/steep/project/dsl.rb
```diff
@@ -74,6 +74,10 @@ def check(*args)
           sources.concat(args)
         end
 
+        def ext(ext)
+          @ext = ext
+        end
```
tk0miya, line 79: Is there any better name for this? I have some other ideas: `file_ext`, `source_ext` and `extensions` (plural).

ParadoxV5: `ext` and `extensions` sound like “plugins”.
Here’s another: `check_ext`

Honestly, it should be called “`templating`”.
ERB templating syntax is not Ruby syntax, and just changing the extension won’t teach Steep how to recognize them, not to mention this design currently blocks a target from continuing checking `*.rb`.
If Steep can automatically switch to ERB mode for `*.erb`s, it can simply change the default file extension to `.{rb,erb}`.

tk0miya: > If Steep can automatically switch to ERB mode for *.erbs, it can simply change the default file extension to .{rb,erb}.

Some template languages, such as HAML, can be converted to Ruby in real-world apps. Therefore it might be better to provide to enhance Steep via plug-ins. From that perspective, it would be better to specify the file extensions via Steepfile.

> P.S. If it’s just for pretending say .ruby files to be synonymous with .rb, is check 'app/views/**/*.{rb,ruby}' sufficient?

Unfortunately, it does not work perfectly.

Please see the diff of `lib/steep/server/master.rb`. It has a hard-coded `.rb` string to set up `didChangeWatchedFiles` event.

PR #1455, opened 2025-01-03
Allow to annotate "self" on toplevel

tk0miya:
To support ERB, this allows to annotate "self" on the top level context. It effects to the whole of the file.

It will also help to support other DSLs (ex. Gemfile, Rakefile, and so on).

Example:

 ```
 # @type self: Gemfile

 group :develop do
   gem 'steep'
   gem 'rbs'
 end
 ```

 ```
 # @type self: Raketask

 namespace :my_group do
   desc "My Task"
   task :my_task do
     ...
   end
 end
 ```

refs: #1409


ParadoxV5:
🤔 should we allow annotating `self` in general?
```rbs
class Example
  …
  def call: () { [self: String?] -> void } -> void
end
```
```ruby
Example.new(…).instance_eval do
  # assert (read: assume) `self` is not nil
  #@type self: String
end
```

tk0miya:
It would be nice to remove the restrictions if there are no reasons.
I also got similar situations when using `.define_method`.

Anyway, your case is not related to this PR. We need to modify another code to remove the restriction of the block's "self".

soutaro (approved):
Thanks!

soutaro:
@ParadoxV5 I believe it already works in blocks, methods, and class/module declarations.

PR #1467, opened 2025-01-14
Fix self union type checking

marcotc:
This fixes the case where a method can return either `self` or another value (e.g. `-> (self | nil)`).

Here's the concrete example: https://github.com/DataDog/dd-trace-rb/blob/5361e8e9a7092ca081ff0d5126daf8950bb2c792/lib/datadog/tracing/span_operation.rb#L202-L223

```ruby
      def stop(stop_time = nil)
        return if stopped?

        now = Core::Utils::Time.now.utc

        start(stop_time || now) unless started?

        @end_time = stop_time || now
        @duration_end = stop_time.nil? ? duration_marker : nil

        events.after_stop.publish(self)

        self
      end
```

Currently this fails with:
```
Cannot allow method body have type `self` because declared as type `(self | nil)`
Diagnostic ID: Ruby::MethodBodyTypeMismatch
```

### DIFF
--- a/lib/steep/subtyping/check.rb
+++ b/lib/steep/subtyping/check.rb
@@ -285,11 +285,6 @@ module Steep
             )
           end
 
-        when relation.sub_type.is_a?(AST::Types::Self) && !self_type.is_a?(AST::Types::Self)
-          Expand(relation) do
-            check_type(Relation.new(sub_type: self_type, super_type: relation.super_type))
-          end
-
         when relation.sub_type.is_a?(AST::Types::Instance) && !instance_type.is_a?(AST::Types::Instance)
           Expand(relation) do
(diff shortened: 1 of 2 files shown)


ParadoxV5:
Could be similar:
```ruby
def example(x) = x
```
```rbs
class Object
  def example: [X < top] (X) -> X?
end
```
```
# Type checking files:

.F

ruby.rb:1:4: [error] Cannot allow method body have type `X` because declared as type `(X | nil)`
│   X <: (X | nil)
│     top <: (X | nil)
│       top <: X
│
│ Diagnostic ID: Ruby::MethodBodyTypeMismatch
│
└ def example(x) = x
      ~~~~~~~
```

soutaro (approved):
Thanks! 👍

PR #1497, opened 2025-02-11
Type narrowing union types via return type of method call

tk0miya:
This tries to narrow types unions by the return type of the method call. If the return type of the method is always truthy or falsy, the receiver type will be narrowed on the truthy branch and the falsy branch.

This helps to narrow types via method definitions (ex. `#nil?`, `#present?`, `#blank?` and so on).

### DIFF
--- a/lib/steep/type_inference/logic_type_interpreter.rb
+++ b/lib/steep/type_inference/logic_type_interpreter.rb
@@ -169,6 +169,14 @@ module Steep
               return [truthy_result, falsy_result]
             end
           else
+            receiver, *_ = node.children
+            receiver_type = typing.type_of(node: receiver) if receiver
+
+            if env[receiver] && receiver_type.is_a?(AST::Types::Union)
+              result = evaluate_union_method_call(node: node, env: env, receiver: receiver, receiver_type: receiver_type)
+              return result if result
+            end
+
             if env[node]
               truthy_type, falsy_type = factory.partition_union(type)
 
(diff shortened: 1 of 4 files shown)


lib/steep/type_inference/logic_type_interpreter.rb
```diff
@@ -408,6 +416,63 @@ def evaluate_method_call(env:, type:, receiver:, arguments:)
...
       end
 
+      def evaluate_union_method_call(node:, env:, receiver:, receiver_type:)
+        call_type = typing.call_of(node: node) rescue nil
+        return unless call_type.is_a?(Steep::TypeInference::MethodCall::Typed)
+
+        truthy_types = [] #: Array[AST::Types::t]
+        falsy_types = [] #: Array[AST::Types::t]
+
+        receiver_type.types.each do |type|
```
tk0miya, line 426: I tried to create a mapping from the types of receivers to the return types of methods for each receivers. Is there a better way to do that?

I used AncestorBuilder and `call_type.method_decls.method_decls[N].method_def.defined_in`. But it is annoying...

lib/steep/type_inference/logic_type_interpreter.rb
```diff
@@ -408,6 +416,63 @@ def evaluate_method_call(env:, type:, receiver:, arguments:)
...
+                  truthy_types << type if truthy
+                  falsy_types << type if falsey
+                  true
+                end
+              end
+            end
+          else
+            truthy_types << type
+            falsy_types << type
+          end
```
tk0miya, line 457: I handled only `Name::*,` `Nil`, and `Literal.` But there are kinds of AST::Types. How do we evaluate them?

test/type_construction_test.rb
```diff
@@ -9944,6 +9944,28 @@ def foo: () -> (Integer | String)
...
+        end
+      RBS
+      source = parse_ruby(<<~RUBY)
+        a = [1].first
+        a + 1 if a.present?
+      RUBY
+
+      with_standard_construction(checker, source) do |construction, typing|
+        construction.synthesize(source.node)
+        assert_no_error typing
```
tk0miya, line 9964: This test verifies that type narrowing of a Union type occurs by confirming that no errors are raised. This verification is somewhat implicit, but is that acceptable? If there is a better testing method, please let me know.

soutaro: This is okay, but I would recommend using `type_check_test.rb` instead.

lib/steep/type_inference/logic_type_interpreter.rb
soutaro: How about using `subtyping.builder.shape(type, config)`?

```rb
          if shape = subtyping.builder.shape(type, config)
            method = shape.methods[call_type.method_name] or raise
            method_type = method.method_types.find do |method_type|
              call_type.method_decls.any? {|decl| subtyping.factory.method_type(decl.method_type) == method_type }
            end
            if method_type
              return_type = method_type.type.return_type
              truthy, falsey = factory.partition_union(return_type)
              truthy_types << type if truthy
              falsy_types << type if falsey
              next
            end
          end

          truthy_types << type
          falsy_types << type
```

Haven't tested the code above carefully, but the test case you wrote works with this.

tk0miya: Cool!!

test/type_construction_test.rb
tk0miya: Moved.

tk0miya:
@soutaro Thank you for reviewing. updated now.

soutaro (approved):
Thanks! 👍

PR #1501, opened 2025-02-19
Implement user-defined type guard method

tk0miya:
This implements the minimal version of user-defined type guard method. A user-defined type guard methos is a method defined by user that is able to guarantee the type of the objects (receiver or arguments).

At present, Steep supports type guard methods provided by ruby-core (ex. `#is_a?`, `#nil?`, and so on).  But we also have many kinds of user-defined methods that are able to check the type of the objects.

Therefore user-defined type guard will help checking the type of these applications by narrowing types.

This implementation uses an annotation to declare user-defined type guard method.

```
class Example < Integer
  %a{guard:self is Integer}
  def integer?: () -> bool
end
```

For example, the following method `Example#integer?` is a user-defined type guard method that narrows the Example object itself to an Integer if the conditional branch passed.

```
example = Example.new
if example.integer?
  example  #=> Integer
end
```

In this PR, the predicate of type guards only supports "self is TYPE" statement.  I have a plan to extend it:

* `%a{guard:self is arg}`
* `%a{guard:self is_a arg}`
* `%a{guard:self is TYPE_PARAM}`
* `%a{guard:arg is TYPE}`

Note: The compatibility of RBS syntax is the large reason of using annotations.  I'm afraid that adding a new syntax to define it will bring breaking change to the RBS, and difficult to use it on common repository or generators (ex. gem_rbs_collection and rbs_rails).

refs:

* https://github.com/ruby/rbs/issues/1822

### DIFF
--- a/lib/steep/interface/builder.rb
+++ b/lib/steep/interface/builder.rb
@@ -281,6 +281,7 @@ module Steep
               overloads = method.defs.map do |type_def|
                 method_name = method_name_for(type_def, name)
                 method_type = factory.method_type(type_def.type)
+                method_type = replace_guard_method(definition, type_def, method_type)
                 method_type = replace_primitive_method(method_name, type_def, method_type)
                 method_type = replace_kernel_class(method_name, type_def, method_type) { AST::Builtin::Class.instance_type }
                 method_type = add_implicitly_returns_nil(type_def.each_annotation, method_type)
@@ -313,6 +314,7 @@ module Steep
               overloads = method.defs.map do |type_def|
                 method_name = method_name_for(type_def, name)
                 method_type = factory.method_type(type_def.type)
+                method_type = replace_guard_method(definition, type_def, method_type)
                 method_type = replace_primitive_method(method_name, type_def, method_type)
                 if type_name.class?
                   method_type = replace_kernel_class(method_name, type_def, method_type) { AST::Types::Name::Singleton.new(name: type_name) }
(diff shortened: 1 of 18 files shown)


lib/steep/type_inference/logic_type_interpreter.rb
```diff
@@ -405,6 +405,38 @@ def evaluate_method_call(env:, type:, receiver:, arguments:)
               truthy_result.update_type { FALSE }
             ]
           end
+
+        when AST::Types::Logic::Guard
+          if receiver
+            receiver_type = factory.deep_expand_alias(typing.type_of(node: receiver)) || raise
+
+            # TODO: Make the type absolute
```
tk0miya, line 413: I'd like to know the best way to absolute the type in the type guard annotation. `factory#absolute_type_name`?

lib/steep/interface/builder.rb
```diff
@@ -725,6 +727,37 @@ def proc_shape(proc, proc_shape)
...
+      end
+
+      def context_from(type_name)
+        if type_name.namespace == RBS::Namespace.root
+          [nil, type_name]
+        else
+          parent = context_from(type_name.namespace.to_type_name)
+          [parent, type_name]
+        end
+      end
```
tk0miya, line 759: I could not find such a helper method in Steep and RBS.
I'm not sure where the best place is.

lib/steep/interface/builder.rb
```diff
@@ -725,6 +727,40 @@ def proc_shape(proc, proc_shape)
...
+        if match
+          subject = match[1] or raise
+          operator = match[2] or raise
+          type_name = match[3] or raise
+
+          type = RBS::Parser.parse_type(type_name)
+          raise "Unknown type: #{type_name}" unless type
+
+          context = context_from(definition.type_name)
+          type = type.map_type_name { factory.absolute_type_name(_1, context: context) or raise "Unknown type: #{_1}" }
```
tk0miya, line 741: Is there a better way to handle errors on Interface::Builder? It seems there is no `typing` object. So `typing.add_error()` can't be used here.

tk0miya: Finally, I implemented these checks in signature/validator.rb.

calebhearth:
I just found this via the [RubyKaigi talk](https://rubykaigi.org/2025/presentations/tk0miya.html). Is there any chance it'll get picked back up?

Further, is there any chance that we could implement it in such a way that Ruby-side classes aren't required? For example,	 the `Published` example in the talk is defined as a module, but I'd be unlikely to define such a module in a real class. Still, it would be nice to narrow the type of e.g., `Article` to `_PublishedArticle` based on something like:

```rbs
interface _PublishedArticle
  attr_reader title: String
end

class Article
  %a{guard:self is _PublishedArticle}
  def published? = true
end
```

PR #1505, opened 2025-02-25
Update types

tk0miya:


sig/steep/diagnostic/ruby.rbs
```diff
@@ -1760,8 +1760,8 @@ module Steep
         # end
         # ```
         #
-        def initialize: (node: Parser::AST::Node, method_name: Symbol, params_pair: [Interface::Function::Params, Interface::Function::Params], result: Subtyping::Result::Base) -> void
-                      | (node: Parser::AST::Node, method_name: Symbol, block_pair: [Interface::Block?, Interface::Block?], result: Subtyping::Result::Base) -> void
+        def initialize: (node: Parser::AST::Node, method_name: Symbol, params_pair: [Interface::Function::Params, Interface::Function::Params], result: Subtyping::Result::t) -> void
+                      | (node: Parser::AST::Node, method_name: Symbol, block_pair: [Interface::Block?, Interface::Block?], result: Subtyping::Result::t) -> void
```
tk0miya, line 1764: I guess using `Result::t` is better than `Result::Base`. Is this correct?

sig/steep/type_inference/context.rbs
```diff
@@ -23,15 +23,15 @@ module Steep
...
+        attr_reader forward_arg_type: [Params, Block?] | true | nil
 
         def initialize: (
           name: Symbol,
           method: RBS::Definition::Method?,
           method_type: Interface::MethodType?,
           return_type: AST::Types::t,
           super_method: RBS::Definition::Method?,
-          forward_arg_type: [Params, Block?]?
+          forward_arg_type: [Params, Block?] | true | nil
```
tk0miya, line 34: It seems `forward_arg_type` should take a true value, right?

PR #1506, opened 2025-02-25
Fix typos

tk0miya:


sig/steep/services/completion_provider.rbs
```diff
@@ -120,7 +120,7 @@ module Steep
 
         attr_reader range: Range
 
-        attr_reader reeiver_type: AST::Types::t
+        attr_reader receiver_type: AST::Types::t
```
tk0miya, line 123: The original implementation is defined as re*c*eiver_type. 
https://github.com/soutaro/steep/blob/944af85e5c6f508456f4a7ab9308a7820f5ea8bd/lib/steep/services/completion_provider.rb#L64

soutaro (approved):
🙏

PR #1507, opened 2025-02-25
Support `InvalidTypeApplication` on `#update_env`

ksss:
ref: https://github.com/ruby/rbs/pull/2289

When `InvalidTypeApplicationError` occurs in the RBS validator, there is a possibility that Steep will raise an **Unexpected error** if the type argument is `RBS::Types::ClassInstance`, `RBS::Types::Interface`, or `RBS::Types::Alias`.

This PR fixes the issue so that problems are reported correctly.


soutaro (approved):
Thanks! 👍

PR #1515, opened 2025-03-08
Dedup error message on ancestors check

ksss:
In ancestors error checking, the same issue may be reported multiple times.  

```
$ cat sample.rbs
class InheritModule < Kernel
end

$ bundle exec steep check
# Type checking files:

F

sample.rbs:1:22: [error] Cannot inherit from a module `::Kernel`
│ Diagnostic ID: RBS::InheritModuleError
│
└ class InheritModule < Kernel
                        ~~~~~~

sample.rbs:1:22: [error] Cannot inherit from a module `::Kernel`
│ Diagnostic ID: RBS::InheritModuleError
│
└ class InheritModule < Kernel
                        ~~~~~~

Detected 2 problems from 1 file
```

This happens because both `#one_instance_ancestors` and `#one_singleton_ancestors` in `RBS::DefinitionBuilder::AncestorBuilder` generate the same error.  

This issue can be reproduced by running the following command in the `smoke/diagnostics-rbs` directory:  

```sh
$ bundle exec steep check --save-expectations=test_expectations.yml
```

So far, this issue seems to occur only in ancestors checking.  

Therefore, I propose limiting the deduplication to this specific case.

### DIFF
--- a/lib/steep/services/signature_service.rb
+++ b/lib/steep/services/signature_service.rb
@@ -294,7 +294,7 @@ module Steep
           end
 
           unless errors.empty?
-            # Builder won't be used.
+            errors.uniq! { |e| [e.class, e.message] }
             factory = AST::Types::Factory.new(builder: RBS::DefinitionBuilder.new(env: env, ancestor_builder: builder))
             return errors.map {|error| Diagnostic::Signature.from_rbs_error(error, factory: factory) }
           end


soutaro (approved):
👍

lib/steep/services/signature_service.rb
```diff
@@ -294,7 +294,7 @@ def update_env(updated_files, paths:)
           end
 
           unless errors.empty?
-            # Builder won't be used.
+            errors.uniq! { |e| [e.class, e.message] }
```
soutaro, line 297: I'm curious if it dedup error message from different classes.

```rbs
class C1 < Kernel end
class C2 < Kernel end
```

ksss: In most cases, `#message` includes location information.  
The reason for not using `#location` is that a few exceptions, such as `RBS::SuperclassMismatchError`, do not implement the `#location` method.

As a result, it ends up as follows, which I think makes sense.

### Before

```
$ bundle exec steep check
# Type checking files:

F

sample.rbs:1:11: [error] Cannot inherit from a module `::Kernel`
│ Diagnostic ID: RBS::InheritModuleError
│
└ class C1 < Kernel end
             ~~~~~~

sample.rbs:1:11: [error] Cannot inherit from a module `::Kernel`
│ Diagnostic ID: RBS::InheritModuleError
│
└ class C1 < Kernel end
             ~~~~~~

sample.rbs:2:11: [error] Cannot inherit from a module `::Kernel`
│ Diagnostic ID: RBS::InheritModuleError
│
└ class C2 < Kernel end
             ~~~~~~

sample.rbs:2:11: [error] Cannot inherit from a module `::Kernel`
│ Diagnostic ID: RBS::InheritModuleError
│
└ class C2 < Kernel end
             ~~~~~~

Detected 4 problems from 1 file
```

### After

```
$ bundle exec steep check
# Type checking files:

F

sample.rbs:1:11: [error] Cannot inherit from a module `::Kernel`
│ Diagnostic ID: RBS::InheritModuleError
│
└ class C1 < Kernel end
             ~~~~~~

sample.rbs:2:11: [error] Cannot inherit from a module `::Kernel`
│ Diagnostic ID: RBS::InheritModuleError
│
└ class C2 < Kernel end
             ~~~~~~

Detected 2 problems from 1 file
```

soutaro: Hmm. It makes sense. 💯

PR #1516, opened 2025-03-08
Implement GitHub formatter

ksss:
I will add a formatter for GitHub Actions.  

By using this formatter in GitHub Actions, it will become easier to display type errors directly in code reviews on github.com.
The default behavior remains unchanged.
Rubocop has a similar feature, and it is also used in [the rbs repository](https://github.com/ruby/rbs/blob/1209f49fd9073374d5660e8d9fc25ce2fc5c1af0/Rakefile#L150).

You can see the demo by https://github.com/ksss/rubocop-on-rbs/pull/86/files also.

image


soutaro (approved):
Awesome! 💯

PR #1517, opened 2025-03-11
Fix union self type refinement based on method call

soutaro:
This PR fixes type checking the following Ruby code:

```
if location = node.location         # Assume `node` has a union type
  # location must be nonnil here
end
```

### DIFF
--- a/lib/steep/type_inference/logic_type_interpreter.rb
+++ b/lib/steep/type_inference/logic_type_interpreter.rb
@@ -173,29 +173,41 @@ module Steep
             receiver_type = typing.type_of(node: receiver) if receiver
 
             if env[receiver] && receiver_type.is_a?(AST::Types::Union)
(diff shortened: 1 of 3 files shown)


lib/steep/type_inference/logic_type_interpreter.rb
```diff
@@ -173,29 +173,41 @@ def evaluate_node(env:, node:, type: typing.type_of(node: node))
             receiver_type = typing.type_of(node: receiver) if receiver
 
             if env[receiver] && receiver_type.is_a?(AST::Types::Union)
-              result = evaluate_union_method_call(node: node, env: env, receiver: receiver, receiver_type: receiver_type)
-              return result if result
```
soutaro, line 177: Returning the result here skipped further variable type refinements.

PR #1670, opened 2025-05-21
Add `RedundantIgnoreComment` diagnostic

sinsoku:
Add a new diagnostic so we can find and remove unnecessary `steep:ignore` comments.


soutaro (approved):
Thank you for the contribution!
I have some things to fix, so merging this first and will fix them myself.

Issue #974, opened 2023-12-06
Typechecking DSL files: top-level context and block context


lloeki:
DSLs are great. Typed DSLs are even better: from the usual type checking of values to leveraging the LSP for autocompletion it would be fantastic that Steep could do a better job.

Here's the situation.

### Steep DSL

Steep has a DSL... so I tried to typecheck `Steepfile` with `steep`!

```
# Steepfile
# @type self: Steep::Project::DSL

target :lib do
  check '.'
  check 'Steepfile'
  check 'Rakefile'

  library 'steep'
  library 'uri'
  library 'rake'

  signature '.'
end
```

Unfortunately, despite the annotation `steep check Steepfile` gives:

```
Steepfile:4:2: [error] Type `::Steep::Project::DSL` does not have method `check`
│ Diagnostic ID: Ruby::NoMethod
│
└   check '.'
    ~~~~~

Steepfile:5:2: [error] Type `::Steep::Project::DSL` does not have method `check`
│ Diagnostic ID: Ruby::NoMethod
│
└   check 'Steepfile'
    ~~~~~

Steepfile:6:2: [error] Type `::Steep::Project::DSL` does not have method `check`
│ Diagnostic ID: Ruby::NoMethod
│
└   check 'Rakefile'
    ~~~~~

Steepfile:8:2: [error] Type `::Steep::Project::DSL` does not have method `library`
│ Diagnostic ID: Ruby::NoMethod
│
└   library 'steep'
    ~~~~~~~

Steepfile:9:2: [error] Type `::Steep::Project::DSL` does not have method `library`
│ Diagnostic ID: Ruby::NoMethod
│
└   library 'uri'
    ~~~~~~~

Steepfile:10:2: [error] Type `::Steep::Project::DSL` does not have method `library`
│ Diagnostic ID: Ruby::NoMethod
│
└   library 'rake'
    ~~~~~~~

Steepfile:12:2: [error] Type `::Steep::Project::DSL` does not have method `signature`
│ Diagnostic ID: Ruby::NoMethod
│
└   signature '.'
    ~~~~~~~~~

Steepfile:3:0: [error] Type `::Object` does not have method `target`
│ Diagnostic ID: Ruby::NoMethod
│
└ target :lib do
  ~~~~~~

Detected 8 problems from 1 file
```

Note that for the last one it tried to look things up in `::Object`

I tried adding more annotation to the block:

```
# @type self: Steep::Project::DSL

target :lib do
[...]

PR #1748, opened 2025-06-21
feat: Add built-in support for Rakefiles

euglena1215:
We use Rake in our work and want to perform type checking on Rakefiles, similar to how we check standard Ruby files.
However, type-checking Rakefiles with the current Steep setup presents several challenges.

First, Rakefiles are not discovered by default.
A `check '.'` directive in the `Steepfile` does not include files with a `.rake` extension, forcing users to explicitly add `check '*.rake'`.

Second, Rake's DSL methods like `namespace` and `task` result in `NoMethodError`s. While a Rakefile's top-level `self` has access to `Rake::DSL` methods, Steep only recognizes it as `main` (an instance of `Object`), leading to incorrect type analysis.

To work around these issues, users currently need to:
1. Explicitly list Rakefile patterns (e.g., `'*.rake'`) in the `Steepfile`.
2. Add a type annotation or ignore comments in each Rakefile: ```rb namespace :foo do # steep:ignore NoMethod # @type self: Object & Rake::DSL task bar: :environment do |name, _args| # do something end end ```

This commit introduces built-in support for Rakefiles by eliminating the need for such manual configuration.
It treats `.rake` files as standard Ruby files by default and automatically types the top-level `self` in Rakefiles as `Object & Rake::DSL`.

### DIFF
--- a/lib/steep/project/dsl.rb
+++ b/lib/steep/project/dsl.rb
@@ -119,15 +119,15 @@ module Steep
         end
 
         def source_pattern
-          Pattern.new(patterns: sources, ignores: ignored_sources, ext: ".rb")
+          Pattern.new(patterns: sources, ignores: ignored_sources, extensions: [".rb", ".rake"])
         end
 
         def signature_pattern
-          Pattern.new(patterns: signatures, ignores: ignored_signatures, ext: ".rbs")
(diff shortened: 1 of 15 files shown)


ParadoxV5:
Half-solves #974; ~~next up: Gemfile, `.gemspec`, Steepfile!~~

PR #1789, opened 2025-07-20
Failing test case.

samuel-williams-shopify:
Steep is not properly refining types after the `||=` (or-assign) operator. When using `obj.attr ||= []` where `attr` is typed as `Array[T] | nil`, Steep should refine the type of `obj.attr` to `Array[T]` after the assignment, but it continues to treat it as `Array[T] | nil`.

I encountered an issue where the following code wouldn't work with steep:

```ruby
(self.values ||= []) << value
```

## Debug Information

Using `steep check --verbose --log-level=debug`, we can see that Steep processes the `or_asgn` operation but doesn't properly refine the type afterward. The debug log shows:

```
[synthesize:(line)] or_asgn
[synthesize:(line)] [synthesize:(line)] or
[synthesize:(line)] check_relation: ... |- nil <: false
[synthesize:(line)] check_relation: ... |- false <: nil
```

This suggests Steep is processing the `||=` but not updating the type environment with the refined type.

## Workaround

The issue can be worked around by explicitly using an if-else statement:

```ruby
if node.values
  node.values << value
else
  node.values = [value]
end
```

This properly triggers type refinement and eliminates the error.

### DIFF
--- a/test/type_construction_test.rb
+++ b/test/type_construction_test.rb
@@ -2186,6 +2186,39 @@ EOF
     end
   end
 
+  def test_or_asgn_type_refinement
+    with_checker(<<~RBS) do |checker|
+        class TestNode
+          attr_reader values(): Array[untyped] | nil
+          def values=: (Array[untyped] | nil) -> Array[untyped]
+        end
+      RBS
+      source = parse_ruby(<<~EOF)
+        node = TestNode.new
+
+        # After this line, node.values should be refined from Array[untyped] | nil to Array[untyped]
+        node.values ||= []
+
+        # This should work without error since ||= guarantees non-nil
+        node.values << "test"
+      EOF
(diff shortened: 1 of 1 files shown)


samuel-williams-shopify:
While experimenting with various fixes, I also found this didn't work:

```
				if @children
					@children << child
				else
					@children = [child]
				end
```

I'm not sure if the expectation is that I write:

```
				if children = @children
					children << child
				else
					@children = [child]
				end
```

soutaro:
Hi @samuel-williams-shopify, unfortunately Steep won't handle this case -- method invocation changes the type of an expression.

https://github.com/soutaro/steep/pull/2205 may improve the situation a little by fixing `(self.values ||= []) << value` typing.

PR #1836, opened 2025-09-08
Convert ERB code into Ruby before type-checking

felixefelip:
ERB to Ruby code conversion is needed to type-check ERB.

Input:
```
<html>
  <body>
    <ul>
    <% languages.each do |langage| %>
      <li><%= language %>
    <% end %>
    </ul>
  </body>
</html>
```

Output: (replace HTML tags with whitespace)
```
      
        
        
       languages.each do |langage|   
              language   
       end   
         
         
       
```

refs: https://github.com/soutaro/steep/issues/1409


felixefelip:
@ParadoxV5 thanks for the review! 

> Good idea, but I suggest making use of [`ERB::Compiler`](https://github.com/ruby/erb/blob/v4.0.4/lib/erb/compiler.rb) so you don’t have to parse things yourself.
> 

The [ERB::Compiler](https://github.com/ruby/erb/blob/v4.0.4/lib/erb/compiler.rb) looks great, I'm going to use it.

> Additionally, ERB isn’t limited to Rails, let alone HTML. Your parser probably doesn’t care, but the test appears designed for Rails **only**.
>

My fault, I'm used ERB only with Rails, I'm going to adjust the tests examples.

> * Supercedes [Steepfile: Allow to change the extension of the source files #1454](https://github.com/soutaro/steep/pull/1454)
> * Do we want to move this to [RBS](https://github.com/ruby/rbs)?

Regarding moving this to RBS, I will analyze the projects better to give an opinion.

felixefelip:
@ParadoxV5 I have not found a way to use [ERB::Compiler](https://github.com/ruby/erb/blob/v4.0.4/lib/erb/compiler.rb) for the proposal of this PR. Could you instruct me on how to use it, please?

Alternatively, I found the [Herb](https://github.com/marcoroth/herb/blob/1f393a2177c9cfca0c4cebc7762cd24e5da56b25/lib/herb/libherb/libherb.rb#L29) that provide the parser we need. It's ok to use that?

ParadoxV5:
@felixefelip:

> Could you instruct me on how to use it, please?

The documentation above `class ERB::Compiler` reads: (formatting mine)

<details>
<summary>expand/collapse section</summary>

> Internally ERB does something like this to generate the code returned by `ERB#src`:
> ```ruby
> compiler = ERB::Compiler.new('<>')
> compiler.pre_cmd    = ["_erbout=+''"]
> compiler.put_cmd    = "_erbout.<<"
> compiler.insert_cmd = "_erbout.<<"
> compiler.post_cmd   = ["_erbout"]
>
> code, enc = compiler.compile("Got <%= obj %>!\n")
> puts code
> ```
> 
> *Generates*:
> ```
>   #coding:UTF-8
>   _erbout=+''; _erbout.<< "Got ".freeze; _erbout.<<(( obj ).to_s); _erbout.<< "!\n".freeze; _erbout
> ```
> 
> By default the output is sent to the `print` method. For example:
> ```ruby
> compiler = ERB::Compiler.new('<>')
> code, enc = compiler.compile("Got <%= obj %>!\n")
> puts code
> ```
> *Generates*:
> ```
> #coding:UTF-8
> print "Got ".freeze; print(( obj ).to_s); print "!\n".freeze
> ```

</details>

That means, `ERB::Compiler.new(flags).compile(erb_file).first` generates a string containing code that ERB executes to process ERB files, with Ruby instructions as-is and ERB text (e.g., HTML) outputted via `Kernel#print`.
You can pass this generated code to some Steep API (or dump to a temporary file if all fails).
You can also optionally configure an `ERB::Compiler` to be more Steep-user-friendly.
(There might be a way to completely ignore the verbatim texts, too.)
```ruby 
erb = ERB::Compiler.new nil
erb.put_cmd = erb.insert_cmd= "\n"
puts erb.compile('Got <%= process(obj) %>!').first
```
```
#coding:UTF-8

 "Got ".freeze;
(( process(obj) ).to_s);
 "!".freeze
```

> It's ok to use that?

Assuming `Herb.extract_ruby` works with non-HTML as well, then my only concern is that ERB is still preferable for its popularity and [StdLib status](https://stdgems.org/erb/).

PR #1845, opened 2025-09-16
Prepare for Ruby 3.4+ syntax support by introducing Prism parser

euglena1215:
To support future Ruby syntax (3.4+), we need to migrate from the parser gem to Prism. 

- Use Prism::Translation::Parser33 instead of Parser::Ruby33
- Add RBS type definitions for Prism parser classes in sig/shims/
- Update test cases to work with both parsers where Prism's stricter checking caused failures, adjusting syntax while preserving original intent

### DIFF
--- a/lib/steep/source.rb
+++ b/lib/steep/source.rb
@@ -19,7 +19,7 @@ module Steep
       @ignores = ignores
     end
 
-    class Builder < ::Parser::Builders::Default
+    class Builder < Prism::Translation::Parser::Builder
       def string_value(token)
         value(token)
       end
--- a/sig/shims/prism.rbs
+++ b/sig/shims/prism.rbs
@@ -1,6 +1,11 @@
 module Prism
   module Translation
-    class Parser33
+    class Parser
+      class Builder < ::Parser::Builders::Default
+      end
+    end
+
+    class Parser33 < Parser
       def initialize: (Parser::Builders::Default builder) -> void
 
       def parse: (Parser::Source::Buffer) -> Parser::AST::Node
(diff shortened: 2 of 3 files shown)


steep.gemspec
```diff
@@ -38,6 +38,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 3.2.0'
 
   spec.add_runtime_dependency "parser", ">= 3.2"
+  spec.add_runtime_dependency "prism", ">= 0.25.0"
```
euglena1215, line 41: It seems that `Prism::Translation::Parser33` has been available since 0.25.0.

lib/steep/source.rb
```diff
@@ -31,9 +31,16 @@ def string_value(token)
     end
 
     def self.new_parser
-      ::Parser::Ruby33.new(Builder.new).tap do |parser|
-        parser.diagnostics.all_errors_are_fatal = true
-        parser.diagnostics.ignore_warnings = true
+      if Gem::Version.new(RUBY_VERSION) <= '3.3'
```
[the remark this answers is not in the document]
euglena1215: After reviewing the prism gem, it appears that the parser in `Prism::Translation` is only available for Ruby 3.3 and later.

While it's debatable whether the condition to use the parser gem should be for Ruby versions less than or equal to 3.3 or simply less than 3.3, some form of conditional logic is necessary given that Steep supports Ruby 3.2.
However, this branching might not be necessary if parsing Ruby 3.2 code with `Prism::Translation::Parser33` is not an issue. (It might be fine, given that all code was already being parsed by `Parser::Ruby33`.)

sig/shims/prism.rbs
```diff
@@ -0,0 +1,13 @@
+module Prism
```
euglena1215, line 1: The `prism` gem has a `sig/` directory, but it was missing the `Prism::Translation::Parser33` class that `steep` uses.

PR #2088, opened 2026-02-21
Fix ancestor error reporting and silent failing

denisenkom:
Fix: Report ancestor errors even when they occur in core library files
When RBS files contain classes that cause ancestor errors (like superclass
mismatches when reopening core modules), Steep would fail to report these
errors because their diagnostic locations pointed to core library files
rather than user code.

The bug occurred in two scenarios:

1. When validating signature files, ancestor errors were filtered by file
   path, hiding errors whose locations pointed to core library files.

2. When building the RBS environment with ancestor errors, the old
   DefinitionBuilder (without user-defined classes) was used instead of
   the new one, causing "Unknown name for build_instance" errors.

Changes:
- lib/steep/services/type_check_service.rb: Report all ancestor errors
  without filtering by path, since they affect the entire environment
- lib/steep/services/signature_service.rb: Use the newly created
  DefinitionBuilder (with user classes) when creating AncestorErrorStatus
- lib/steep/diagnostic/signature.rb: Prefer user file locations over
  core library locations when reporting superclass mismatch errors
- Add smoke test to prevent regression

Fixes the issue where type checking would silently fail when reopening
core modules like Errno with different superclass hierarchies.

### DIFF
--- a/lib/steep/diagnostic/signature.rb
+++ b/lib/steep/diagnostic/signature.rb
@@ -593,9 +593,27 @@ module Steep
             location: error.location
           )
         when RBS::SuperclassMismatchError
+          # Try to find a declaration in a user file (not in core/stdlib)
+          # If the type is reopened in user code, use that location instead of the core library location
+          location = error.entry.primary_decl.location
+          
+          # Check if there are other declarations in non-core files
+          # context_decls is an array of [context, decl] pairs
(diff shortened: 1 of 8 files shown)


soutaro:
@denisenkom Thank you so much for opening this PR. I've fixed the issue myself in #2198. 🙇‍♂️

Issue #2095, opened 2026-02-26
Steep 1.10.0 causes Ruby::MethodBodyTypeMismatch, Steep 1.9.4 did not.


znz:
`typechars` in [my add-rbs branch of bitclust](https://github.com/znz/bitclust/tree/add-rbs) causes `Ruby::MethodBodyTypeMismatch` after steep updated to 1.10.0.

How to reproduce:

```
% cat Gemfile
# frozen_string_literal: true

source "https://rubygems.org"

gem "steep", "~> 1.9.0"
% cat m.rb
module M
  def typechars(type, mpattern)
    if type                     then [type]
    elsif /\A[A-Z]/ =~ mpattern then [':', '.#']
    else                             ['.#', ':']
    end
  end
end
% cat m.rbs
module M
  def typechars: (String? type, String? mpattern) -> ([String] | [":", ".#"] | [".#", ":"])
end
% bundle exec steep check
2026-02-26 14:05:44.297: ERROR: [Steep 1.9.4] Cannot find a configuration at Steepfile: `steep init` to scaffold. Using current directory...
# Type checking files:

..

No type error detected. 🧋
% vi Gemfile
% cat Gemfile
# frozen_string_literal: true

source "https://rubygems.org"

gem "steep"
% bundle update --all
Fetching gem metadata from https://rubygems.org/.............
Resolving dependencies...
Using unicode-display_width 3.2.0 (was 2.6.0)
Using terminal-table 4.0.0 (was 3.0.2)
Using steep 1.10.0 (was 1.9.4)
Bundle updated!
% bundle exec steep check
2026-02-26 14:06:14.938: ERROR: [Steep 1.10.0] Cannot find a configuration at Steepfile: `steep init` to scaffold. Using current directory...
# Type checking files:

F.

m.rb:2:6: [error] Cannot allow method body have type `([::String] | [::String, ::String])` because declared as type `([::String] | [":", ".#"] | [".#", ":"])`
│   ([::String] | [::String, ::String]) <: ([::String] | [":", ".#"] | [".#", ":"])
│     [::String, ::String] <: ([::String] | [":", ".#"] | [".#", ":"])
│       [::String, ::String] <: [::String]
│
│ Diagnostic ID: Ruby::MethodBodyTypeMismatch
│
└   def typechars(type, mpattern)
        ~~~~~~~~~

Detected 1 problem from 1 file
```

PR #2096, opened 2026-02-26
Fix array literal type inference with union tuple hint

rhiroe:
resolve https://github.com/soutaro/steep/issues/2095

## Summary

When an array literal is type-checked against a hint that is a union of tuple types (e.g. `["yes", "ok"] | ["no", "error"]`), the type checker was incorrectly inferring `[::String, ::String]` instead of the correct specific tuple type.

## Root Cause

`#try_tuple_type` returns a `Pair` even when the inferred type does not match the given tuple hint. The first non-nil pair was accepted unconditionally, so `["no", "error"]` tried against hint `["yes", "ok"]` produced `[::String, ::String]`, which was accepted instead of trying the next hint `["no", "error"]`.

This behavior was introduced in #359 (More tuple types through hint), which deliberately removed the subtype check to support type variable hints like `[A, B]` (issue #328). The unconditional acceptance was a side effect.

## Fix

Reintroduce subtype validation against each individual tuple in the union. If a candidate passes the check, return it immediately. If no candidate passes, fall back to the first candidate rather than exiting the tuple-matching path entirely.

**Why the fallback is necessary:** without it, a mismatched annotation like `[1, ""] #: [1, "", bool]` would silently fall through to `try_array_type` and infer `[::Integer, ::String]` without reporting any error. The fallback ensures the mismatch is still surfaced as a `FalseAssertion` diagnostic.

**Why issue #328 is not regressed:** for type variable hints like `[A, B]`, `check_relation([::Integer, ::String] <: [A, B])` succeeds, so the `return` path is taken directly and the fallback is never needed.

### DIFF
--- a/lib/steep/type_construction.rb
+++ b/lib/steep/type_construction.rb
@@ -1762,13 +1762,18 @@ module Steep
               if hint
                 tuples = select_flatten_types(hint) {|type| type.is_a?(AST::Types::Tuple) } #: Array[AST::Types::Tuple]
                 unless tuples.empty?
+                  fallback_pair = nil #: Pair?
                   tuples.each do |tuple|
                     typing.new_child() do |child_typing|
                       if pair = with_new_typing(child_typing).try_tuple_type(node, tuple)
-                        return pair.with(constr: pair.constr.save_typing)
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks! 🙏

Issue #1574, opened 2025-04-11
Steep fails to find files to typecheck when running from within a directory that contains curly braces in its path


lorenyu:
When running `steep` from within a directory that contains curly braces in its name, it fails to find files to type check.

## Reproduction Steps:
I created a repository https://github.com/navapbc/ruby-steep-bug-repro that reproduces this error, based on the example provided in the [Getting started with Steep in 5 minutes](https://github.com/soutaro/steep/blob/master/guides/src/getting-started/getting-started.md) page.

In the repo I included two examples, one that works and one that doesn't. I created two identical directories named `foo` and `{{foo}}`, each containing a `lib` folder with an example `hello.rb` file that has a single type error. Running `steep check` from within the `foo` directory successfully finds the type error, while running it from within the `{{foo}}` directory fails to find any files to type check.

### Running from within `foo` works

```bash
➜ cd foo && bundle exec steep check
# Type checking files:

F

lib/hello.rb:4:39: [error] Cannot pass a value of type `::String` as an argument of type `::Symbol`
│   ::String <: ::Symbol
│     ::Object <: ::Symbol
│       ::BasicObject <: ::Symbol
│
│ Diagnostic ID: Ruby::ArgumentTypeMismatch
│
└ puts "Hello! The price is #{currencies[country]}100. 💸"
                                         ~~~~~~~

Detected 1 problem from 1 file
```

### Running from within `foo` doesn't work

```bash
➜ cd \{\{foo\}\} && bundle exec steep check

# Type checking files:



No type error detected. 🧉
```

PR #2203, opened 2026-04-10
Handle special characters in path when globbing

soutaro:
Copy of https://github.com/soutaro/steep/pull/1575


soutaro:
Closes https://github.com/soutaro/steep/issues/1574

PR #2229, opened 2026-05-24
feat(postconditions): read intrinsic literal RHS type for narrowing detection

felixefelip:
Closes #34.

`Postconditions::Inferrer` was missing setter narrowings whenever the ivar was already declared in RBS. The `:ivasgn` synthesize passes the declared type as `hint:` to the RHS synthesize, and `test_literal_type` returns the hint itself when the literal is compatible — so:

\`\`\`ruby
attr_accessor :name #: String?

def set_default_name
  @name = "TBA Venue"     # str node types as String?, not String
end
\`\`\`

The "strictly narrower than declared" check fails (`String? == String?`), and no postcondition is emitted. Cross-receiver narrowing breaks end-to-end whenever the target class has prior RBS.

## What changed

`intrinsic_type_of(node)` in `Postconditions::Inferrer` — for literal RHS nodes (`:str`, `:int`, `:sym`, `:nil`, `:true`/`:false`, `:float`, `:regexp`, `:dstr`, `:dsym`) computes the type directly from the node shape (`AST::Builtin::String.instance_type` for `:str`, etc.), matching what `synthesize(node, hint: nil)` would return without the widening. Non-literal RHS (sends, lvars, arrays, hashes) fall back to the typed-out `type_of` — those rarely suffer the widening since the hint mostly affects literal value-class lookups.

`collect_ivar_refinements` now calls `intrinsic_type_of` instead of `type_of`.

## Why scoped to the Inferrer

The widening at `:ivasgn` synthesize is intentional Steep behavior. Collections need it:

\`\`\`ruby
# @x: Array[Numeric]
@x = [1, 2, 3]
# sem hint → Array[Integer], que NÃO é subtype de Array[Numeric] (Array é invariante)
# com hint → Array[Numeric]
\`\`\`

Changing the synthesize itself would be a behavior change for all Steep users. Doing the intrinsic-type computation only inside the Inferrer keeps the fix local and reversible. Issue #34 carries an upstream-friendly path (option (2): expose `IntrinsicType.of(node)` as a public helper) for future follow-up.

## Test plan

- [x] `test_infers_narrowing_when_rhs_is_string_literal_against_nilable_ivar` — the actual bug, pins the new behavior.
- [x] `test_infers_narrowing_when_rhs_is_nil_literal_against_nilable_ivar` — already worked, pinned to catch `:nil` regressions.
- [x] Full Steep suite: 1493 runs, 1 pre-existing failure (`test_inline_module_alias_basic`), 0 new regressions.

## Pipeline impact

This is the last piece needed to close the order_factory `Concerts::TicketMounter#mount_ticket_event(venue)` end-to-end (for the literal-narrowing portion; the delegation-chain piece is still in steep#32). After this PR, `venue.set_default_name` followed by `venue.name.upcase` typechecks clean — `set_default_name` finally emits an `unconditional.self: "::Venue & ::Venue::AfterSetDefaultName"` postcondition, and rbs_infer-generated marker class makes the receiver-type intersection resolve.

Note: rbs_infer's `SteepBridge#ivar_write_types_per_method` has the same widening issue on its own narrowing-detection path (separate from this PR). A matching fix on the rbs_infer side will follow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


felixefelip:
Apologies — opened in the wrong repo. Re-opening on the fork (felixefelip/steep) since this is fork-scoped work for now.

PR #2237, opened 2026-06-09
Add the %a{warning: ...} annotation in

sampersand:
This PR adds in support for the `%a{warning: ...}` annotation, which can be attached to constants, methods, or global variables.

## Motivating Usage

This is the usage that motivated it (from https://github.com/ruby/rbs/pull/2987), slightly simplified:

```rbs
class Array[E]
  def min: () { (E a, E b) -> Comparable::_CompareToZero } -> E?
         | # ... other branches ...
end
```
This is the correct signature—the return value of the block _cannot_ return `nil`. However, most `<=>` implementations _do_ return `nil`, and having to do a kludge (eg `array.min { |a,b| (a.foo <=> b.foo) || raise("never happens") }`) is quite un-ergonomic, awkward, and can potentially break backwards-compat.

The solution is to add in a warning annotation:
```rbs
class Array[E]
  def min: () { (E a, E b) -> Comparable::_CompareToZero } -> E?
         | %a{warning: returning nil from comparisons always raises} () { (E a, E b) -> Comparable::_CompareToZero? } -> E?
         | # ... other branches ...
end
```

### DIFF
--- a/lib/steep/annotations_helper.rb
+++ b/lib/steep/annotations_helper.rb
@@ -15,38 +15,58 @@ module Steep
       nil
     end
 
-    def deprecated_type_name?(type_name, env)
-      annotations =
+    def warning_annotation?(annotations)
+      annotations.reverse_each do |annotation|
+        if match = annotation.string.match(/\Awarning(:\s*(?<message>.+))?\z/)
+          return [annotation, match[:message]]
+        end
+        if match = annotation.string.match(/\Asteep:warning(:\s*(?<message>.+))?\z/)
+          return [annotation, match[:message]]
+        end
+      end
+
+      nil
+    end
+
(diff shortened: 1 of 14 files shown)


lib/steep/annotations_helper.rb
```diff
@@ -15,39 +15,59 @@ def deprecated_annotation?(annotations)
       nil
     end
 
-    def deprecated_type_name?(type_name, env)
-      annotations =
+    def warning_annotation?(annotations)
```
soutaro, line 18: Can you extract the warning annotation test to `RBS::AST::Annotation#warning?`? so that we can reuse it or clear explanations to the RBS users.

(Also for `RBS::AST::Annotation#deprecation?`)

PR #2238, opened 2026-06-13
feat: support the `it` block parameter

### LINKED ISSUE #1450: Support `it` (read: It is time to switch to Prism.)
Late Merry Christmas! Ruby 3.4 is out and–
```console
$ cat ruby.rb; steep check
5.then { it }
# Type checking files:

.F

ruby.rb:1:9: [error] Type `::Object` does not have method `it`
│ Diagnostic ID: Ruby::NoMethod
│
└ 5.then { it }
           ~~

```

The problem:
```ruby
irb(main):001> require 'parser/current'
warning: parser/current is loading parser/ruby34, which recognizes 3.4.0-compliant syntax, but you are running 3.4.1.
Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
=> true
irb(main):002> Parser::CurrentRuby.parse('5.then { it }')
=>
s(:block,
  s(:send,
    s(:int, 5), :then),
  s(:args),
  s(:send, nil, :it))
irb(main):003> Parser::CurrentRuby.parse('5.then { _1 }') # comparison
=>
s(:numblock,
  s(:send,
    s(:int, 5), :then), 1,
  s(:lvar, :_1))
```

But should we bother the Parser team?
* whitequark/parser#1046

With the conclusion of [the Syntax Moratorium](https://bugs.ruby-lang.org/issues/20575), who know what ~~pipeline operators~~ creative sugar the devs of Ruby will come up with.
Combined with syntax *bug fixes*, Prism the “official” parser will make maintainers new and old lose interest of older alternatives, sooner or later.
```ruby
irb(main):001> Prism.parse '5.then { it }'
=>
              └── block:
                  @ BlockNode (location: (1,7)-(1,13))
                  ├── flags: ∅
                  ├── locals: []
                  ├── parameters:
                  │   @ ItParametersNode (location: (1,7)-(1,13))
                  │   └── flags: ∅
                  ├── body:
…
```

I can bare with sticking with `_1`s for the time beïng.

tk0miya:
Recognize Ruby 3.4's `it` (e.g. `[1].map { it + 1 }`) by handling the new :itblock node alongside :numblock across type checking, source mapping, diagnostics, hover, and signature help.

This requires bumping Prism::Translation::Parser33 to Parser34.

Closes #1450

### DIFF
--- a/lib/steep/source.rb
+++ b/lib/steep/source.rb
@@ -31,7 +31,7 @@ module Steep
     end
 
     def self.new_parser
-      Prism::Translation::Parser33.new(Builder.new).tap do |parser|
+      Prism::Translation::Parser34.new(Builder.new).tap do |parser|
         parser.diagnostics.all_errors_are_fatal = true
         parser.diagnostics.ignore_warnings = true
       end
@@ -257,7 +257,7 @@ module Steep
         annot.line or next
 
         case node.type
-        when :def, :module, :class, :block, :numblock, :ensure, :defs, :resbody
+        when :def, :module, :class, :block, :numblock, :itblock, :ensure, :defs, :resbody
           location = node.loc
           location.line <= annot.line && annot.line < location.last_line
         else
@@ -565,13 +565,13 @@ module Steep
                   end
                 ]
               )
-            when :numblock
-              send, size, body = node.children
+            when :numblock, :itblock
+              send, arg, body = node.children
               node = node.updated(
                 nil,
                 [
                   map_child_node(send) {|child| insert_type_node(child, child_assertions) },
-                  size,
+                  arg,
                   insert_type_node(body, child_assertions)
                 ]
               )
@@ -648,7 +648,7 @@ module Steep
         case node.type
         when :send, :csend
           node
-        when :block, :numblock
+        when :block, :numblock, :itblock
           send = node.children[0]
           case send.type
           when :send, :csend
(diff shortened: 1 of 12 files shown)


lib/steep/source.rb
```diff
@@ -31,7 +31,7 @@ def string_value(token)
     end
 
     def self.new_parser
-      Prism::Translation::Parser33.new(Builder.new).tap do |parser|
+      Prism::Translation::Parser34.new(Builder.new).tap do |parser|
```
tk0miya, line 34: Discussion: I bumped the Prism parser from Parser33 to Parser34 to support the `it` parameter.
But this introduces incompatible behavior.

```ruby
def it
  1
end
[1, 2, 3].map { it }
```

With Parser33, the `it` inside the block is recognized as a method call to `#it`. With Parser34, the same `it` is recognized as the block's implicit parameter.

I'm afraid that this brings a breaking change. Would it be better to add a configuration for the Ruby syntax version?

tk0miya:
@soutaro Is there any chance to merge this? This is a blocker to migrating my code to the new syntax.

lib/steep/source.rb
soutaro: I don't think it's a problem.
Ruby 3.3 was released 2023. Two years old. And it's already in security maintenance stage. We may want to use even newer version --`Parser40`.

tk0miya:
Thanks! I can't wait for the next release :-)

PR #2253, opened 2026-07-30
Infer tuple types for array literals against pair interface hints like `Hash::_Pair`

sferik:
RBS 4.1 changed the block types of `Array#to_h` and `Hash#to_h`, and the array-of-pairs argument of `Hash.[]`, from tuples to the new `Hash::_Pair[K, V]` interface (https://github.com/ruby/rbs/pull/2694). Steep only derives tuple hints from hint types that are tuple types themselves, so a pair literal like `[key, value]` in a `to_h` block is now inferred as an array and every call of the form `to_h { |x| [key, value] }` fails with `Ruby::BlockBodyTypeMismatch`:

```
Cannot allow block body have type `::Array[(::Symbol | ::Integer)]` because declared as type `::Hash::_Pair[::Symbol, ::Integer]`
```

`Hash::_Pair` is not a tuple type but it declares the tuple it accepts through its `#to_ary` method. When the hint contains no tuple type directly, this change deconstructs interface and intersection hints with `try_convert(type, :to_ary)` and uses the resulting tuple types as hints. The array literal is then inferred as a tuple and the existing tuple validation and fallback machinery applies unchanged. All three `_Pair` call sites funnel through the same array-literal hint path, so one change covers them.

The intersection case arises when the block body hint combines a tuple with the interface, for example `([::Symbol, ::Numeric] & ::Hash::_Pair[::Symbol, ::Numeric])`. Flattening finds no tuple there because the intersection is opaque to `flatten_union`, but the shape of the intersection still answers `#to_ary` with a tuple.

A genuinely mismatched pair body still fails with `Ruby::BlockBodyTypeMismatch`, and the diagnostic improves, reporting the precise tuple type of the body instead of an array type. A regression test covers this.

### DIFF
--- a/lib/steep/type_construction.rb
+++ b/lib/steep/type_construction.rb
@@ -1761,6 +1761,19 @@ module Steep
             else
               if hint
                 tuples = select_flatten_types(hint) {|type| type.is_a?(AST::Types::Tuple) } #: Array[AST::Types::Tuple]
+                if tuples.empty?
+                  # An interface hint like `Hash::_Pair[K, V]` (which rbs 4.1 uses for the blocks of `Array#to_h`
+                  # and `Hash#to_h` and for the array-of-pairs argument of `Hash.[]`) is not a tuple type itself,
+                  # but declares the tuple it accepts through its `#to_ary` method.
+                  # Deconstruct such hints so the array literal can be inferred as a tuple.
+                  candidates = select_flatten_types(hint) do |type|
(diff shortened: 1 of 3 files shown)


soutaro:
Thank you for the PR! 🙏 The same inference landed on master in 81cb48cf as part of #2267 (RBS 4.2.0.pre.1 update). I've rebased the branch onto master, keeping your regression tests.

PR #2255, opened 2026-08-03
Fix self-type substitution for reopened generic modules

### LINKED ISSUE #2254: `Enumerable` self-type constraint fails with RBS 4.1
After upgrading `rbs` from 4.0.3 to 4.1.2, Steep reports a `RBS::ModuleSelfTypeError` for a class that includes a typed `Enumerable` and defines a matching `each` method.

The same signature passes with RBS 4.0.3. Replacing the block parameter with `untyped` also passes, but unnecessarily weakens the public signature.

#### Versions

- Ruby: 4.0.6
- Steep: 2.0.0
- Passing RBS version: 4.0.3
- Failing RBS version: 4.1.2

#### Minimal reproduction

```rbs
class Item
end

class Collection
  include Enumerable[Item]

  def each: () { (Item) -> void } -> void
end
```

The original signature also has a no-block overload, but the error reproduces without it:

```rbs
def each: () { (Item) -> void } -> void
        | () -> Enumerator[Item, void]
```

Run:

```sh
bundle exec steep check
```

#### Expected behavior

The signature should satisfy `Enumerable[Item]`'s `_Each` self-type constraint. The block passed to `each` receives exactly the declared enumerable element type, `Item`.

#### Actual behavior

With RBS 4.1.2, Steep reports:

```text
# Type checking files:

F.

sig/repro.rbs:5:2: [error] Module self type constraint in type `::Collection` doesn't satisfy: `::Collection <: ::_Each[Elem, void]`
│   ::Collection <: ::_Each[Elem, void]
│     #<Steep::Interface::Shape: type=::Collection, private?=false, methods={!, !=, !~, <=>, ==, ===, DelegateClass, Digest, __id__, __send__, all?, any?, chain, chunk, chunk_while, class, clone, collect, collect_concat, compact, count, cycle, define_singleton_method, detect, display, drop, drop_while, dup, each, each_cons, each_entry, each_slice, each_with_index, each_with_object, entries, enum_for, eql?, equal?, extend, filter, filter_map, find, find_all, find_index, first, flat_map, freeze, frozen?, grep, grep_v, group_by, hash, include?, inject, inspect, instance_eval, instance_exec, instance_of?, instance_variable_defined?, instance_variable_get, instance_variable_set, instance_variables, is_a?, itself, kind_of?, lazy, map, max, max_by, member?, method, methods, min, min_by, minmax, minmax_by, nil?, none?, object_id, one?, partition, private_methods, protected_methods, public_method, public_methods, public_send, reduce, reject, remove_instance_variable, respond_to?, reverse_each, select, send, singleton_class, singleton_method, singleton_methods, slice_after, slice_before, slice_when, sort, sort_by, sum, take, take_while, tally, tap, then, to_a, to_enum, to_h, to_json, to_s, to_set, to_yaml, uniq, yield_self, zip} <: #<Steep::Interface::Shape: type=::_Each[Elem, void], private?=false, methods={each}
│       { () { (::Item) -> void } -> void } <: { () { (Elem) -> void } -> void }
│         { () { (::Item) -> void } -> void } <: () { (Elem) -> void } -> void
│           () { (::Item) -> void } -> void <: () { (Elem) -> void } -> void
│             { (::Item) -> void } <: { (Elem) -> void }
│               (Elem) -> void <: (::Item) -> void
│                 (Elem) <: (::Item)
│
│ Diagnostic ID: RBS::ModuleSelfTypeError
│
└   include Enumerable[Item]
    ~~~~~~~~~~~~~~~~~~~~~~~~

```

The diagnostic appears to retain `Elem` as an unresolved type variable instead of substituting the `Item` argument from `Enumerable[Item]`.

#### Workarounds investigated

Using `untyped` for the block parameter passes:

```rbs
def each: () { (untyped) -> void } -> void
```

However, this loses block-parameter checking for direct calls to `each`.

Making `each` generic can satisfy the module constraint:

```rbs
def each: [Elem < Item] () { (Elem) -> void } -> void
```

That signature is not sound for this implementation. The collection can yield any `Item`, while the supplied block could accept only a narrower subtype.

Steep correctly reports a block type mismatch when such a block is forwarded to an `Array[Item]#each` implementation.

Pinning RBS to the 4.0 line preserves the proper concrete signature:

```ruby
gem "rbs", "~> 4.0.3"
```

#### Additional context

RBS 4.1 defines `Enumerable` with an `_Each` self type resembling:

```rbs
interface _Each[out T, out R = void]
  def each: () { (T) -> void } -> R
end

module Enumerable[unchecked out E] : _Each[E]
```

Is this unresolved self-type substitution a Steep compatibility issue with the RBS 4.1 definition, or should this be reported to RBS instead?

#### Repo script

<details>
<summary><code>steep-rbs-enumerable-repro.rb</code></summary>

```ruby
# frozen_string_literal: true

require "bundler/inline"
require "fileutils"
require "tmpdir"

rbs_version = ARGV.fetch(0, "4.1.2")

gemfile(true) do
  source "https://rubygems.org"

  gem "rbs", rbs_version
  gem "steep", "2.0.0"
end

require "steep"
require "steep/cli"

Dir.mktmpdir("steep-rbs-repro") do |directory|
  FileUtils.mkdir_p(File.join(directory, "lib"))
  FileUtils.mkdir_p(File.join(directory, "sig"))

  File.write(File.join(directory, "Steepfile"), <<~RUBY)
    target :repro do
      check "lib"
      signature "sig"

      library "prism"
      library "rbs"
    end
  RUBY

  File.write(File.join(directory, "lib/repro.rb"), <<~RUBY)
    class Item
    end

    class Collection
      include Enumerable

      def each
        yield Item.new
      end
    end
  RUBY

  File.write(File.join(directory, "sig/repro.rbs"), <<~RBS)
    class Item
    end

    class Collection
      include Enumerable[Item]

      def each: () { (Item) -> void } -> void
    end
  RBS

  puts "\nReproducing with Steep 2.0.0 and RBS #{rbs_version}\n\n"

  status = Dir.chdir(directory) do
    Steep::CLI.new(
      argv: ["check"],
      stdout: $stdout,
      stderr: $stderr,
      stdin: $stdin,
    ).run
  end

  exit(status)
end
```

</details>

Reproduces the failure:
```bash
ruby steep-rbs-enumerable-repro.rb
```

Passes for comparison:
```bash
ruby steep-rbs-enumerable-repro.rb 4.0.3
```

marcoroth:
This pull request fixes module self-type validation when a generic module is reopened with different type parameter names.

RBS treats the type parameters of reopened declarations positionally, so declarations using `E` and `Elem` are equivalent. 

Steep previously substituted every self type using the parameter names from the module's primary declaration. A self type originating from another declaration could therefore retain an unresolved type variable and incorrectly produce `RBS::ModuleSelfTypeError`.

The signature validator now finds the declaration that introduced each self type and uses that declaration's parameter names for substitution. A regression test covers equivalent module reopenings with differently named parameters:

```rbs
module Enum[E] : _Each[E]
end

module Enum[Elem] : _Each[Elem]
end

class Collection
  include Enum[String]

  def each: () { (String) -> void } -> void
end
```

This also allows the RBS 4.1.2 core and shim declarations for `Enumerable` to be validated together.

Resolves #2254

### DIFF
--- a/lib/steep/signature/validator.rb
+++ b/lib/steep/signature/validator.rb
@@ -214,7 +214,17 @@ module Steep
           ancestor_ancestors.self_types or raise
           ancestor_ancestors.params or raise
           self_constraints = ancestor_ancestors.self_types.map do |self_ancestor|
-            s = Interface::Substitution.build(ancestor_ancestors.params, args)
+            params = ancestor_ancestors.params
+
+            if source = self_ancestor.source
+              if entry = env.class_decls[ancestor.name]
+                if decl = entry.each_decl.find { _1.is_a?(RBS::AST::Declarations::Module) && _1.self_types.include?(source) }
+                  params = decl.type_params.map(&:name)
+                end
+              end
+            end
+
+            s = Interface::Substitution.build(params, args)
             ancestor_to_type(self_ancestor).subst(s)
           end
 
--- a/sig/test/validation_test.rbs
+++ b/sig/test/validation_test.rbs
@@ -41,6 +41,8 @@ class ValidationTest < Minitest::Test
 
   def test_validate_mixin_class: () -> untyped
 
+  def test_validate_mixin_with_renamed_type_params: () -> untyped
+
   def test_validate_mixin_module: () -> untyped
 
   def test_validate_instance_variables: () -> untyped
--- a/test/validation_test.rb
+++ b/test/validation_test.rb
@@ -485,6 +485,33 @@ end
     end
   end
 
+  def test_validate_mixin_with_renamed_type_params
+    with_checker <<~RBS do |checker|
+      interface _Each[A]
+        def each: () { (A) -> void } -> void
+      end
+
+      module Enum[E] : _Each[E]
+      end
+
+      module Enum[Elem] : _Each[Elem]
+      end
+
+      class Collection
+        include Enum[String]
+
+        def each: () { (String) -> void } -> void
+      end
+    RBS
+
+      Validator.new(checker: checker).tap do |validator|
+        validator.validate_one_class(RBS::TypeName.parse("::Collection"))
+
+        assert_predicate validator, :no_error?
+      end
+    end
+  end
+
   def test_validate_mixin_module
     with_checker <<-EOF do |checker|
 interface _FooEach[A]


soutaro:
Thanks for opening this PR. 🙏 

This problem is actually fixed in RBS 4.2, and [this PR](https://github.com/soutaro/steep/pull/2277) requires RBS `~> 4.2`.

PR #2261, opened 2026-08-13
Accept block-pass arguments for optional blocks

### LINKED ISSUE #1207: Symbol#to_proc passed as an optional block argument causes a warning "Cannot pass a value of type `::Proc` as a block-pass-argument"
This issue seems to be similar to #709 and #652, but slightly different. This is a pretty annoying problem when using Rubocop as it prefers Symbol#to_proc over blocks.

```sh
> steep --version
1.7.1
```

Consider following code:
```ruby
module Test
  def pass
    yield 1
  end

  def with_block
    pass { -_1 }
  end

  def with_to_proc
    pass(&:-@)
  end

  def with_proc
    pass(&->(v) { -v } )
  end
end
```
with following signatures:
```rbs
module Test
  def pass: () ?{ (Integer value) -> Integer } -> Integer

  def with_block: () -> Integer
  def with_to_proc: () -> Integer
  def with_proc: () -> Integer
end
```
Only the `&:-@` construct is getting a warning:
```
lib/test.rb:11:9: [warning] Cannot pass a value of type `::Proc` as a block-pass-argument of type `(^(::Integer) -> ::Integer | nil)`
│   ::Proc <: (^(::Integer) -> ::Integer | nil)
│     ::Proc <: ^(::Integer) -> ::Integer
│
│ Diagnostic ID: Ruby::BlockTypeMismatch
│
└     pass(&:-@)
           ~~~~
```

However, changing the block to be non-optional suddenly gets rid of the warning:
```rbs
def pass: () { (Integer value) -> Integer } -> Integer
```
It seems that the case of just passing a `#to_proc` value is handled, but optionality of the block causes something to break.

soutaro:
Fixes #1207.

`&:sym` and `&method(:name)` fail against an **optional** block:

```
test.rb:1:13: [error] Cannot pass a value of type `::Proc` as a block-pass-argument of type `(^(::Integer) -> void | nil)`
│   ::Proc <: (^(::Integer) -> void | nil)
│     ::Proc <: ^(::Integer) -> void
│
│ Diagnostic ID: Ruby::BlockTypeMismatch
│
└ Foo.new.each(&:to_s)
               ~~~~~~
```

The `Symbol#to_proc` / `Method#to_proc` special cases in `:block_pass` match on `AST::Types::Proc`, but the hint of a block-pass argument is `^(...) -> ... | nil` when the block is optional, so neither runs and the argument keeps the plain `::Proc` it converts to.

Fixed by hinting the argument with the block type itself. `BlockPassArg` builds both, so it exposes `proc_type` next to `node_type`, and the union stays where it belongs — the subtyping check that lets `nil` through for an optional block.

This removes the need for ruby/rbs#3060.

### DIFF
--- a/lib/steep/type_construction.rb
+++ b/lib/steep/type_construction.rb
@@ -4358,7 +4358,9 @@ module Steep
             when arg.compatible?
               if arg.node
                 # Block pass (&block) is given
-                node_type, constr = constr.synthesize(arg.node, hint: arg.node_type)
+                # Passing `proc_type` instead of `node_type` because a block-pass argument is a proc, not `nil`.
+                # `nil` is still allowed by the `node_type` check below.
+                node_type, constr = constr.synthesize(arg.node, hint: arg.proc_type)
 
                 nil_given =
                   constr.check_relation(sub_type: node_type, super_type: AST::Builtin.nil_type).success? &&
--- a/lib/steep/type_inference/send_args.rb
+++ b/lib/steep/type_inference/send_args.rb
@@ -473,16 +473,20 @@ module Steep
           end
         end
 
+        def proc_type
+          raise unless block
+
+          AST::Types::Proc.new(type: block.type, block: nil, self_type: block.self_type)
+        end
+
         def node_type
           raise unless block
 
-          type = AST::Types::Proc.new(type: block.type, block: nil, self_type: block.self_type)
-
           if block.optional?
-            type = AST::Types::Union.build(types: [type, AST::Builtin.nil_type])
+            AST::Types::Union.build(types: [proc_type, AST::Builtin.nil_type])
+          else
+            proc_type
           end
-
-          type
         end
       end
 
--- a/sig/steep/type_inference/send_args.rbs
+++ b/sig/steep/type_inference/send_args.rbs
@@ -199,6 +199,14 @@ module Steep
 
         def pair: () -> [Parser::AST::Node, Interface::Function]?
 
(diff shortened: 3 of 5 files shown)


lib/steep/type_construction.rb
```diff
@@ -5030,6 +5032,20 @@ def try_tuple_type(array_node, hint)
       constr.add_typing(array_node, type: AST::Types::Tuple.new(types: element_types))
     end
 
+    # Returns the proc type a block-pass argument is expected to have
```
soutaro: Delete the comments here since it's duplicated with the docs in the RBS file.

test/type_construction_test.rb
```diff
@@ -6975,6 +6975,74 @@ def m: [T] () { (untyped) -> T } -> T
     end
   end
 
+  def test_block_pass_and_symbol__optional_block
```
soutaro: Prefer TypeCheckTest with this test. Add tests here if it cannot be tested with the end-to-end test.

soutaro: Moved to `TypeCheckTest` — all three cases turned out to be reachable end-to-end, so nothing is left here. `test_block_pass_optional_block` covers `&:sym` and `&method(:name)` (alongside the no-block, literal-block and lambda forms), and `test_block_pass_optional_block_mismatch` pins the `Ruby::BlockTypeMismatch` that must still be reported.

---
_Generated by [Claude Code](https://claude.ai/code)_

